PR #7, opened 2020-04-15
Put copies number

cthulhu-rider:


object/service.proto
```diff
@@ -84,6 +84,8 @@ message PutRequest {
         Object Object       = 1;
         // Token with session public key and user's signature
         session.Token Token = 2;
+        // Number of the object copies to store within the RPC call (zero is processed according to the placement rules)
+        uint32 CopiesNumber = 3;
```
im-kulikov, line 88: mb we can be named it like in the container service? `ReplFactor`

realloc: The meaning is different here. Technically it's a number of ACKs from Storage nodes. Let it be Copies here to simplify understanding.

im-kulikov: thanks for explaining

cthulhu-rider: @im-kulikov, explanations from @realloc were ahead of mine.
@realloc, clarify, please, do you agree w/ `CopiesNumber` or propose to trim it to `Copies`.

realloc: `CopiesNumber` is fine

PR #8, opened 2020-04-20
Add gRPC method to change current node state

im-kulikov:
The main objective of this method is to allow to control and manage the state of the node by its owner.

- if the user is not declared in the settings, the system returns an access error.
- if the new status is not defined or is not known, the system notifies about that.
- if the new status is known to the system, it will be applied and the node will change the current status.

### DIFF
--- a/state/service.proto
+++ b/state/service.proto
@@ -10,7 +10,8 @@ import "github.com/gogo/protobuf/gogoproto/gogo.proto";
 
 option (gogoproto.stable_marshaler_all) = true;
 
-// Status service provides node's healthcheck and status info
+// Status service provides node's healthcheck and status info.
+// TODO: decide how to describe auth and were contains permissions.
 service Status {
     // Netmap request allows to receive current [bootstrap.SpreadMap](bootstrap.md#bootstrap.SpreadMap)
     rpc Netmap(NetmapRequest) returns (bootstrap.SpreadMap);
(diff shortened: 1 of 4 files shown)


state/service.proto
```diff
@@ -95,3 +99,27 @@ message DumpVarsRequest {
 message DumpVarsResponse {
     bytes Variables = 1;
 }
+
+// ChangeStateRequest message allows to change current node state of node.
+// To permit access, used server config options.
+// The request should be signed.
```
cthulhu-rider, line 105: I suggest leaving here only a comment on the body of the message (a brief description of the fields, the need for a signature).

state/service.proto
```diff
@@ -27,6 +27,10 @@ service Status {
     // To permit access, used server config options.
     // The request should be signed.
     rpc DumpVars(DumpVarsRequest) returns (DumpVarsResponse);
+    // ChangeState allows to change current node state of node.
+    // To permit access, used server config options.
+    // The request should be signed.
```
cthulhu-rider, line 32: I suggest leaving here only the call destination (what can be done)

im-kulikov: There’s description, that used in an other calls

state/service.proto
im-kulikov: Yeah, I’ll remove last 2 lines

im-kulikov: What you think about “.. contains a new node state”?

cthulhu-rider: Sounds laconic and correct

state/service.proto
cthulhu-rider: Oh, i see now.

state/service.proto
im-kulikov: done

state/service.proto
im-kulikov: added TODO to decide later

PR #9, opened 2020-04-24
Changes for updated session token

cthulhu-rider:
Summary:

- replace `Raw` field from `object.GetRequest` and `object.HeadRequest` to `service.RequestMetaHeader`;

- remove `Origin` field from `service.RequestMetaHeader_Signature` message;

- replace `session.Token` message to service package;

- replace `object.Header_Verify` of type `session.VerificationHeader` with `object.Header_Token` of type `service.Token`;

- remove no longer used `session.VerificationHeader` message;

- update `service.Token` message structure;

- add `service.Token` message to `service.RequestVerificationHeader`.


object/service.proto
```diff
@@ -83,7 +80,7 @@ message PutRequest {
         // Object with at least container id and owner id fields
         Object Object       = 1;
         // Token with session public key and user's signature
-        session.Token Token = 2;
+        service.Token Token = 2;
```
AlexVanin, line 83: Why `SearchRequest`, `GetRangeRequest`, `GetRangeHashRequest` don't have token field?

cthulhu-rider: They never had before. In the next changes, `Token` field will be removed from the custom fields of all requests due to the presence in `RequestVerificationHeader`.

object/types.proto
```diff
@@ -36,8 +36,8 @@ message Header {
         Transform Transform                    = 4;
         // Tombstone header that set up in deleted objects
         Tombstone Tombstone                    = 5;
-        // Verify header that contains session public key and user's signature
-        session.VerificationHeader Verify      = 6;
+        // Token header that contains session token
```
AlexVanin, line 39: I would like to specify in comments, that token header contains session token, that was used during object creation.

object/service.proto
AlexVanin: Got it.

AlexVanin:
Just in case, are we merging these token requests into master branch? @realloc

object/types.proto
cthulhu-rider: Yes, it would clarify the actual meaning. Done.

PR #11, opened 2020-04-27
service: remove gogoproto customname from embedded Token.Info field

cthulhu-rider:
This commit:

  * removes gogoproto customname option from embedded Token.Info field;

  * renames Info field of Token message to TokenInfo.


im-kulikov:
Mb you talk about ‘customname’?

PR #17, opened 2020-06-18
service: add Bearer token message to RequestVerificationHeader

cthulhu-rider:


service/verify.proto
```diff
@@ -91,3 +94,26 @@ message TokenLifetime {
 //     uint32 Version = 2;
 //     bytes Data = 3;
 // }
+
+// BearerTokenMsg carries information about request ACL rules with limited lifetime
+message BearerTokenMsg {
+    message Info {
+        // ACLRules is an ACL rules represented in bytes
```
AlexVanin, line 101: I would rather have description similar to `EACL` field in container: `carries binary representation of the table of extended ACL rules`. Or at least mention, that it is *extended* ACL rules.

PR #27, opened 2020-08-04
object: Change the structure of Object protobuf message

### LINKED ISSUE #26: Change the structure of Object protobuf message
In [latest](https://github.com/nspcc-dev/neofs-api/releases/tag/v1.2.0) neofs-api version [Object](https://github.com/nspcc-dev/neofs-api/blob/b6a03439192df84dbe8b58e1ca7625e0e73c8ee1/object/types.proto#L122) structure consists of:

- System header;
- Extended headers;
- Payload.

However, the list of extended headers also includes system information.
In turn, extended headers are defined through protobuf [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) which complicates the declaration of a strict header structure.

Proposed changes:
- unify `System` and `Extended` headers in a single `Header` message;
- make `Object` to be `Header` and `Payload` pair;
- rename `System` header to `Main`;
- move all fields of the `System` header except `ID`, `CID`, `OwnerID` and `PayloadLength` to `Extended` header;
- replace `CreationPoint` message with numerical `CreationEpoch`;
- rename `UserHeader` to object `Attribute`;
- encapsulate the headers of the object splitting hierarchy in the `Split` header.

cthulhu-rider:
closes #26


refs/types.proto
```diff
@@ -12,8 +12,14 @@ option (gogoproto.goproto_stringer_all) = false;
 
 // Address of object (container id + object id)
 message Address {
-    // ObjectID is an object identifier, valid UUIDv4 represented in bytes
-    bytes ObjectID = 1[(gogoproto.customtype) = "ObjectID", (gogoproto.nullable) = false];
+    // ObjectID carries object identifier.
+    ObjectID ObjectID = 1;
```
realloc, line 16: Should we change OID/CID order in address? Having CID first might be more understandable.

object/types.proto
```diff
@@ -1,134 +1,128 @@
...
 
-message Range {
-    // Offset of the data range
-    uint64 Offset = 1;
-    // Length of the data range
-    uint64 Length = 2;
-}
+    // Main carries the main part of the header.
+    // Main MUST NOT be NULL.
+    Main main = 1;
```
realloc, line 29: Does letter case matter here?

object/types.proto
```diff
@@ -1,134 +1,128 @@
...
+        }
 
-message IntegrityHeader {
-    // HeadersChecksum is a checksum of all above headers in the object
-    bytes HeadersChecksum   = 1;
-    // ChecksumSignature is an user's signature of checksum to verify if it is correct
-    bytes ChecksumSignature = 2;
-}
+        // Integrity carries object integrity evidence.
+        Integrity integrity = 1;
```
realloc, line 64: Does letter case matter here?

object/types.proto
```diff
@@ -1,134 +1,128 @@
...
+
+        // CreationEpoch carries number of NeoFS epoch on which the object was created.
+        uint64 CreationEpoch = 3;
+
+        // Tombstone groups the options for deleting an object.
+        message Tombstone {
+        }
+
+        // Tombstone marks the object to be deleted.
+        Tombstone tombstone = 4;
```
realloc, line 86: Does letter case matter here?

object/types.proto
```diff
@@ -1,134 +1,128 @@
...
+
+            // Children carries list of identifiers of the objects generated by splitting the current.
+            repeated refs.ObjectID Children = 4;
+
+            // Origin carries the header of the origin object.
+            Header Origin = 5;
+        }
+
+        // Split carries the position of the object in the split hierarchy.
+        Split split = 7;
```
realloc, line 113: Does letter case matter here?

object/types.proto
```diff
@@ -1,134 +1,128 @@
...
     }
-    // Type of object transformation
-    Type type  = 1;
+
+    // Extended carries the additional part of the header.
+    Extended extended = 2;
```
realloc, line 117: Does letter case matter here?

refs/types.proto
cthulhu-rider: Yes, I planned to switch the field order in the coming API changes.

object/types.proto
cthulhu-rider: Yes, `Main` field name declaration causes compilation error. However, such a message field is generated by the compiler in the public field of the structure. I researched this issue a bit and found this approach valid according to the [docs](https://developers.google.com/protocol-buffers/docs/proto3#nested).

PR #28, opened 2020-08-05
storagegroup: Сhange the structure and storage method

cthulhu-rider:
### Structure
In the previous version, the lifetime of a storage group was measured in NeoFS and UNIX epochs. Since operational work with UNIX epochs is not supported on the system, storage group lifetime needs to be measured in NeoFS system epochs only.

### Storage
In previous version, information about storage group was stored in the dedicated field of extended object header. This provoked redundancy of data in the header, since this field was only used to check for existence. The object payload is best suited as a storage location for information about a storage group. At the same time, to distinguish objects with such a specialized payload, it is necessary to mark their header.

## Summary
To support proposed changes, there is a need to:

- convert `Lifetime` field of the `storagegroup.StorageGroup` message to NeoFS epoch number type;

- change the type of `StorageGroup` field of the `object.Header.Extended` message to a blank marker.


object/types.proto
```diff
@@ -7,7 +7,6 @@ option csharp_namespace = "NeoFS.API.Object";
 
 import "refs/types.proto";
 import "service/verify.proto";
-import "storagegroup/types.proto";
```
realloc, line 10: :+1:

PR #29, opened 2020-08-05
Changes for API v2

cthulhu-rider:


container/types.proto
```diff
@@ -11,23 +13,28 @@ import "netmap/types.proto";
 // and access control information. ID of the container is a 32 byte long
 // SHA256 hash of stable-marshalled container message.
 message Container {
-  // OwnerID carries identifier of the container owner.
+    // OwnerID carries identifier of the container owner.
   refs.OwnerID OwnerID = 1;
```
alexvanin, line 17: This field has two whitespaces instead of four.

container/types.proto
```diff
@@ -11,23 +13,28 @@ import "netmap/types.proto";
...
+    message Attribute {
+        // Key of immutable container attribute.
+        string Key = 1;
+
+        // Value of immutable container attribute.
+        string Value = 2;
+    }
+
+    // Attributes define any immutable characteristics of container.
   repeated Attribute Attributes = 4;
```
alexvanin, line 36: This field has two whitespaces instead of four.

PR #40, opened 2020-08-11
Session token context

cthulhu-rider:
## Proposal
In the current version of the format, a session token is issued to a specific address of an object, which creates an inseparable connection between sessions and object operations. To unify the session mechanism, it is required to abstract the token from the object.
## Summary
Change `Token.Info.address` field of type `refs.Address` with `oneof` enumeration of session contexts. In the initial version, only the object context is available. Additional contexts can be added without losing backward compatibility.


realloc:
What if we rename `Token` to `SessionToken` right here?

Issue #38, opened 2020-08-07
Change ObjectID format from UUID to Hash


realloc:
We need to switch from using UUID as ObjectID to having ObjectID as a commutative Hash of Payload and Headers.

Having UUID as ObejctID was continent at project's early stages, but in NeoFS objects are immutable and we need to reflect that in addressing scheme.

ObjectID should be moved out of Header structure to Object message top level. This also affects the way we create Tombstones and split objects.

PR #43, opened 2020-08-11
[#38] Change ObjectID format from UUID to Hash

realloc:
NeoFS Object are now Content-addressed. 

It means the Object's address depends on it's content. ObjectID is now calculated as hash of Header, which contains a hash of payload. If either if Object's payload of Headers change, the ID will also change.


object/service.proto
```diff
@@ -222,10 +224,25 @@ message HeadRequest {
...
+      // Type of the object payload content
+      ObjectType object_type = 4;
+      // Size of payload in bytes.
+      // 0xFFFFFFFFFFFFFFFF means `payload_length` is unknown
+      uint64 payload_length = 5;
+    }
+    // Carries the requested object header or it's part
+    oneof Head{
+      Header header = 1;
+      MainHeader main_header = 2;
```
alexvanin, line 243: This `oneof` looks very confusing. I would like to see something like this

```proto
    oneof Head {
       Header header = 1;
       ShortHeader short_header = 2;
    }
```

object/types.proto
```diff
@@ -8,126 +8,77 @@ option csharp_namespace = "NeoFS.API.Object";
...
+
+  // Information about spawning the objects through a payload splitting.
+  message Split {
+    // Identifier of the origin object.
+    // Parent and children objects must be within the same container.
+    // Parent object_id id known only to the minor child.
+    refs.ObjectID parent = 1;
+    // Previous carries identifier of the left split neighbor.
+    refs.ObjectID previous = 2;
+    // `signature` field of the parent object. Used to reconstruct parent.
```
alexvanin, line 60: Isn't it signature of `parent_header` then?

object/types.proto
```diff
@@ -8,126 +8,77 @@ option csharp_namespace = "NeoFS.API.Object";
...
-      // with cut creator_key and checksum_signature.
-      bytes header_checksum = 2;
-
-      // session_token carries token of the session within which the object was created.
-      // If session token is presented in object, it acts as the user's proof of the
-      // correctness of the creator_key.
-      service.SessionToken session_token = 3;
-
-      // creator_key carries public key of the object creator in a binary format.
-      bytes creator_key = 4;
```
alexvanin, line 56: This implementation misses key of owner id.

object/types.proto
realloc: Technically it is, in the same was as ObjectID.

object/service.proto
```diff
@@ -153,17 +159,16 @@ message PutResponse {
   // authenticate the nodes of the message route and check the correctness
   // of transmission.
   service.ResponseVerificationHeader verify_header = 3;
-}
+}    // Groups initialization parameters of object placement in NeoFS.
```
alexvanin, line 162: Is this intentional comment line?

realloc: Typo. Fixed =)

PR #44, opened 2020-08-11
Add Matryoshka-style request and response headers

realloc:
For the cases when we need to relay a request between several nodes, we need to have a mechanism to record message path in a verifiable manner. The proposed solution is to recursively add signed request/response meta information on each hop.

### DIFF
--- a/service/meta.proto
+++ b/service/meta.proto
@@ -6,6 +6,7 @@ option go_package = "github.com/nspcc-dev/neofs-api-go/service";
 option csharp_namespace = "NeoFS.API.Service";
 
 import "acl/types.proto";
+import "refs/types.proto";
 import "service/verify.proto";
 
 message XHeader {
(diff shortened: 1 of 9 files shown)


service/verify.proto
```diff
@@ -15,17 +15,17 @@ message Signature {
...
 }
 
-// RequestVerificationHeader is a set of signatures of every NeoFS Node that
-// processed request.
+// Verification info for request signed by all intermediate nodes
 message RequestVerificationHeader {
-  // Signatures is a set of signatures of every passed NeoFS Node
-  repeated Signature signatures = 1;
+  Signature body_signature = 1;
+  Signature meta_signature = 2;
```
alexvanin, line 24: Is `origin_signature` redundant here?

service/verify.proto
```diff
@@ -112,3 +112,11 @@ message BearerTokenMsg {
   // Signature is a signature of token information
   bytes signature = 3;
 }
+
+// Verification info for response signed by all intermediate nodes
+message ResponseVerificationHeader {
+  Signature body_signature = 1;
+  Signature meta_signature = 2;
```
alexvanin, line 35: Is `origin_signature` redundant here?

service/meta.proto
```diff
@@ -13,19 +24,106 @@ message Version {
...
+    // OwnerID carries identifier of the token owner
+    refs.OwnerID owner_id = 2;
+    // Token expiration and valid time period parameters
+    TokenLifetime lifetime = 3;
   }
+  // Bearer Token body
+  Body token = 1;
+
+  // Signature of BearerToken body
+  Signature signature = 3;
```
alexvanin, line 94: `Signature signature = 2;`

PR #45, opened 2020-08-13
Add buf linter call to Makefile

realloc:
Added `buf` linter call to `Makefile` and made soe fixes to make it silent.


accounting/service.proto
```diff
@@ -1,17 +1,17 @@
 syntax = "proto3";
 
-package accounting;
+package neo.fs.accounting.v2;
```
alexvanin, line 3: Don't you want to use `neo.fs.v2.package-name`? This will match `neofs-api-go` import structure.

realloc: typo. fixed.

object/types.proto
```diff
@@ -12,34 +12,46 @@ import "service/verify.proto";
...
-  REGULAR = 0;
+  OBJECT_TYPE_UNSPECIFIED = 0;
+
   // Used internally to identify deleted objects
-  TOMBSTONE = 1;
+  OBJECT_TYPE_TOMBSTONE = 1;
+
   // Identifies that the object holds StorageGroup information
-  STORAGE_GROUP = 2;
+  OBJECT_TYPE_STORAGE_GROUP = 2;
```
alexvanin, line 21: Autogenerated code have enum prefix in these constant. `ObjectType_OBJECT_TYPE_STORAGE_GROUP` will look ugly.

refs/types.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
 
-package refs;
+package neo.fs.v2.refs;
 
-option go_package = "github.com/nspcc-dev/neofs-api-go/refs";
-option csharp_namespace = "NeoFS.API.Refs";
+option go_package = "github.com/nspcc-dev/neofs-api-go/v2/refs";
```
alexvanin, line 5: `option go_package = "github.com/nspcc-dev/neofs-api-go/v2/refs;refs"` ?

service/meta.proto
```diff
@@ -28,67 +31,87 @@ message Version {
...
-      OBJECT_DELETE = 4;
+      VERB_DELETE = 5;
+
       // Refers to object.GetRange RPC call
-      OBJECT_RANGE = 5;
+      VERB_RANGE = 6;
+
       // Refers to object.GetRangeHash RPC call
-      OBJECT_RANGEHASH = 6;
+      VERB_RANGEHASH = 7;
```
alexvanin, line 76: `OBJECT` prefix considered that this enum can be expanded for non object related tokens, therefore non object related verbs. I think it was good idea.

storagegroup/types.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
 
-package storagegroup;
+package neo.fs.v2.storagegroup;
 
-option go_package = "github.com/nspcc-dev/neofs-api-go/storagegroup";
-option csharp_namespace = "NeoFS.API.StorageGroup";
+option go_package = "github.com/nspcc-dev/neofs-api-go/v2/storagegroup";
```
alexvanin, line 5: `option go_package = "github.com/nspcc-dev/neofs-api-go/v2/storagegroup;storagegroup"`?

realloc: typo. fixed.

refs/types.proto
realloc: typo.fixed.

object/types.proto
realloc: Agree

service/meta.proto
realloc: I'd propose to separate contexts by type, i.e. have different context structures for each service type. Maybe we should keep verb enum as is for now and change it in the following PRs.

object/types.proto
realloc: Agree. Changed for all enums except default zero-valued first UNSPECIFIED fields. They must be unique package-wide.

PR #46, opened 2020-08-13
Add per-service context to SessionToken

realloc:
If we allow all types of requests to be forwarded between nodes, it would require additional context included in SessionToken. For now we have only ObjectService request forwarding, but in future we may add more.

### DIFF
--- a/service/meta.proto
+++ b/service/meta.proto
@@ -39,6 +39,41 @@ message TokenLifetime {
   uint64 iat = 3;
 }
 
+// Context information for Session Tokens related to ObjectService requests
+message ObjectServiceContext {
+  // Object request verbs
+  enum Verb {
+    // Unknown verb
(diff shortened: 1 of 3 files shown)


service/meta.proto
```diff
@@ -39,55 +39,94 @@ message TokenLifetime {
...
+  // Verb is a type of request for which the token is issued
+  Verb verb = 1;
+  // Related Object address
+  neo.fs.v2.refs.Address address = 2;
+}
 
-      // Refers to object.GetRange RPC call
-      OBJECT_RANGE = 6;
+// Context information for Session Tokens related to Service requests
+message ContainerServiceContext {
```
alexvanin, line 77: I would rather not add `ContainerServiceContext` until we figure out how neofs-node treats tokens for container requests. We can add it later without compatibility breaks.

PR #47, opened 2020-08-14
Fix go_option in netmap package

alexvanin:


realloc:
Should this also be applied to all other packages?

PR #50, opened 2020-08-14
Add version fields where it makes sense

realloc:
Some pieces of data (container, object, etc) may be stored for a long time and
there will be a need in the future to understand which obscure format from the
past was used to create it.


alexvanin:
`NodeInfo` structure stored in blockchain as a network map, so it may be useful to have version there too.

realloc:
Network map is formed each epoch and in case of structure format change it can be updated by active nodes. Inactive nodes will be removed from network map, so it looks like the NodeInfo structure can always be of the latest version, hence maybe there is no need for a version field.

PR #53, opened 2020-08-18
Fixing types placemnt

realloc:
Minor fixes to use common types and declare them in a common way.


object/service.proto
```diff
@@ -181,10 +180,7 @@ message DeleteRequest {
   message Body {
     // Carries the address of the object to be deleted.
     neo.fs.v2.refs.Address address = 1;
-
-    // Carries identifier the object owner.
-    neo.fs.v2.refs.OwnerID owner_id = 2;
-  }
+ }
```
alexvanin, line 183: Don't we replace `owner_id` with `signature` similar to container delete request?

realloc: Won't it be redundant in Object case? We don't need to pass it to Smart Contract and the deletion evidence in tombstone will have the full request anyway.

PR #58, opened 2020-08-23
Simplify Placement Policy

realloc:
Proposed protobuf format for Placement Policy simplifies direct editing in visual editors and import/export from other formats like JSON.

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -5,89 +5,95 @@ package neo.fs.v2.netmap;
 option go_package = "github.com/nspcc-dev/neofs-api-go/v2/netmap/grpc;netmap";
 option csharp_namespace = "NeoFS.API.v2.Netmap";
 
+// Operations on filters
+enum Operation {
+  // No Operation defined
+  OPERATION_UNSPECIFIED = 0;
+
+  // Equal
+  EQ = 1;
+
+  // Not Equal
+  NE = 2;
+
+  // Greater then
+  GT = 3;
+
+  // Greater or equal
+  GE = 4;
+
+  // Less then
+  LT = 5;
+
+  // Less or equal
+  LE = 6;
+
+  // Logical OR
+  OR = 7;
+
+  // Logical AND
+  AND = 8;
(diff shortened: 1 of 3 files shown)


netmap/types.proto
```diff
@@ -5,89 +5,96 @@ package neo.fs.v2.netmap;
...
+  string key = 2;
+
+  // Filtering operation
+  Operation op = 3;
+
+  // Value to match
+  string value = 4;
+
+  // Next Filter logical relation
+  Operation next_op = 5;
```
fyrchik, line 54: Shouldn't this be also a `Filter`? What value does this operation applies to?

realloc: That's exactly not to use Filter here and have chain of Filters without recursion.

fyrchik: Does the semantics of `(a AND) (b OR) (c AND) (d *)` correspond to the DNF `(a AND b) OR (c AND d)` or to the `((a AND b) OR c) AND d`?

realloc:
Latest PR update proposes to have nested filters, but only top level filter can be named.

Example in JSON:
```json
{
  "name": "BigDataInJapan",
  "op": "AND",
  "filters": [
    {
      "op": "OR",
      "filters": [
        {
          "key": "PowerSupply",
          "op": "GT",
          "value": "9000"
        },
        {
          "key": "BigDataStorage",
          "op": "EQ",
          "value": "true"
        },
      ]
    },
    {
      "key": "Country",
      "op": "EQ",
      "value": "Japan"
    }
  ]
}
```

Same higher level Operation applies to all lower level filters.

Example in JSON:
```json
  {
    "name": "CatStor",
    "op": "OR",
    "filters": [
      {
        "key": "StorageType",
        "op": "EQ",
        "value": "SSD"
      },
      {
        "name": "GoodOldEurope"
      },
      {
        "name": "RAMDisk"
      }      
    ]
  }
```

alexvanin (approved):
Seems good and clean.

PR #67, opened 2020-09-04
Use tuple of full header and signature at object head response body

alexvanin:
Object.Head method returns either full header or short header. Since it can't return anything else, signature must be paired with full header in explicit tuple message.


cthulhu-rider (approved):
LGTM

realloc (approved):

PR #72, opened 2020-10-13
Update API documentation

realloc:
Update NeoFS API auto-generated documentation. It's automatically includes in NeoFS Technical Specification.


netmap/types.proto
```diff
@@ -121,7 +121,39 @@ message NodeInfo {
...
+  // Node's attributes are mostly used during Storage Policy evaluation to
+  // calculate object's placement and find a set of nodes satisfying policy
+  // requirements. There are some "well-known" node attributes common to all the
+  // Storage Nodes in the network and used implicitly with default values if not
+  // explicitly set:
+  //
+  // * Capacity \
+  //   Total available disk space in Gigabytes.
+  // * Price \
+  //   Price in GAS tokens for storing one GB of data during one Epoch.
```
alexvanin, line 135: Just a short note: this could be a small fraction of a number, but network map operates with `uint64` price values. We have to be careful with it because it is not the value that will be directly stored in network map.

accounting/service.proto
```diff
@@ -9,25 +9,24 @@ import "accounting/types.proto";
...
-//
-// To gain access to the requested information, the request body must be formed
-// according to the requirements from the system specification.
+// BalanceRequest message
 message BalanceRequest {
-  //Request body
+  // To indicate the account for which the balance is requested, it's identifier
+  // is used. It can be any existing account in NeoFS sidechain `Balance` smart
+  // contract. If omitted, client implementation MUST set it to the request's
+  // signer `OwnerID`.
```
alexvanin, line 26: I am not sure if ` If omitted, client implementation MUST set it to the request's signer OwnerID` note is necessary in protocol description. By body definition, request does not suggest you to omit this field.

accounting/service.proto
```diff
@@ -9,25 +9,24 @@ import "accounting/types.proto";
 import "refs/types.proto";
 import "session/types.proto";
 
-// The service provides methods for obtaining information
-// about the account balance in NeoFS system.
+// Accounting service provides methods for interaction with NeoFS sidechain via
+// other NeoFS nodes to get information about the account balance. Deposit and
+// Withdraw operations can't be implemented here, as they require Mainnet NeoFS
+// Smart Contract invocation.
```
alexvanin, line 15: `Smart Contract` in lower case

alexvanin (approved):

accounting/service.proto
realloc: Both protobuf encoding and NeoFS API technically allow it to be omitted or set to zero. Added this note to emphasize the expected default behaviour.

im-kulikov:
@realloc can we add some specifications about ID encoding/decoding?

1. `Encoding` to a string: `ID -> bytes -> base58 -> string`
2. `Decoding` from a string: `string -> base58 -> bytes -> ID`

realloc:
> @realloc can we add some specifications about ID encoding/decoding?
> 
>     1. `Encoding` to a string: `ID -> bytes -> base58 -> string`
> 
>     2. `Decoding` from a string: `string -> base58 -> bytes -> ID`

Sure, but it will be added to `neofs-spec` in API v2 Intro section.

im-kulikov:
> > @realloc can we add some specifications about ID encoding/decoding?
> > ```
> > 1. `Encoding` to a string: `ID -> bytes -> base58 -> string`
> > 
> > 2. `Decoding` from a string: `string -> base58 -> bytes -> ID`
> > ```
> 
> Sure, but it will be added to `neofs-spec` in API v2 Intro section.

thanks!

acl/types.proto
```diff
@@ -86,69 +88,81 @@ enum HeaderType {
...
 
-  // Action carries ACL target action.
+  // Rule execution result. Either allows or denies access if filters match.
   Action action = 2 [json_name = "Action"];
 
-  // Filter definition
+  // Filter to check particular properties of the request or object.
   message Filter {
-    // Header carries type of header.
+    // Define if Object ot Request header will be used
```
alexvanin, line 101: Object of Request

netmap/types.proto
```diff
@@ -121,16 +129,51 @@ message NodeInfo {
...
     // Key of the node attribute.
     string key = 1;
 
     // Value of the node attribute.
     string value = 2;
 
-    // Parent keys, if any
-    // Example: For City it can be Region or Country
+    // Parent keys, if any. For example for `City` it could be `Region` and
+    // `Country`.
```
alexvanin, line 176: I think only `Country` can be parent for the `City` bucket. It stores direct predecessors, not the whole branch.

container/service.proto
```diff
@@ -272,12 +286,14 @@ message GetExtendedACLRequest {
 
 // Get Extended ACL
 message GetExtendedACLResponse {
-  // Response body
+  // Get Extended ACL Response body can be empty if the requested container did
+  // not have Extended ACL Table attached or Extended ACL was not allowed at
+  // container creation.
```
alexvanin, line 291: > or Extended ACL was not allowed at container creation.

Is there any ways to do that, actually? Container contract does not checks container fields such as basic ACL to make such restriction. You can set extended ACL but it will have no power.

container/service.proto
```diff
@@ -10,46 +10,51 @@ import "container/types.proto";
...
+// nodes.
 service ContainerService {
-  // Put invokes 'Put' method in container smart-contract and returns
-  // response immediately. After new block in morph chain, request is verified
-  // by inner ring nodes. After one more block in morph chain, container
-  // added into smart-contract storage.
+  // `Put` invokes `Container` smart contract's `Put` method and returns
+  // response immediately. After a new block is issued in sidechain, request is
+  // verified by Inner Ring nodes. After one more block in sidechain, container
+  // is added into smart contract storage.
```
alexvanin, line 21: I am not sure but maybe it is worth to mention, that smart contract transfers internal assets to the inner ring nods therefore it is payed operation.

netmap/types.proto
realloc: St.Petersburg looks at you in disbelief.

container/service.proto
realloc: Agree. I meant that it will be ignored if not allowed in `BasicACL`. Will fix.

netmap/types.proto
alexvanin: I mean in default case of `Region -> Country -> City` all buckets except first one will have only one parent.

realloc: Maybe we should use another example here. Geo locations are covered with Locode now. Do you have any proposal? Like `DC->Room->Rack` or availability zones?

alexvanin: > Geo locations are covered with Locode now.

Yes, they are. However inner ring nodes will set country, city and region attributes anyways. Are we building network map bucket graph like this one?
```
     Region
      |  \
      |   Country
      |   /
      City   
```

`DC->Room->Rack` should work too as an example.

alexvanin (approved):

netmap/types.proto
realloc: This should be valid. Let's try to look at it in a different issue.

Issue #73, opened 2020-10-15
Add JSON field names where applicable


realloc:
Most of the NeoFS API structures can be serialized into JSON, but not all structures have corresponding `json_name` tags assigned.

`json_name` tags should be added and serialization procedure specified at least for:

* accounting.Decimal
* acl.BearerToken
* container.Container
* netmap.NodeInfo
* netmap.PlacementPolicy
* object.Object
* storagegroup.StorageGroup

And related structures

PR #74, opened 2020-10-16
[#73] Add json field names to oblect struct

fyrchik:
Related #73.


alexvanin:
Seems like protobuf linter does not support json mapping.

realloc:
> Seems like protobuf linter does not support json mapping.

It does. It actually works with AST-like structure. However, when not set explicitly, the JSON mapping names are automatically transformed from filed names. In the check it gives errors because of that field name changes break compatibility.

Think, we can ignore this at initial mapping addition.

object/types.proto
```diff
@@ -32,53 +32,53 @@ enum MatchType {
...
+  uint64 creation_epoch = 4 [json_name = "CreationEpoch"];
 
   // Size of payload in bytes.
   // `0xFFFFFFFFFFFFFFFF` means `payload_length` is unknown
-  uint64 payload_length = 5;
+  uint64 payload_length = 5 [json_name = "PayloadSize"];
 
   // Hash of payload bytes
-  neo.fs.v2.refs.Checksum payload_hash = 6;
+  neo.fs.v2.refs.Checksum payload_hash = 6  [json_name = "PayloadHash"];
```
alexvanin, line 71: Two spaces between 6 and [

object/types.proto
```diff
@@ -32,53 +32,53 @@ enum MatchType {
...
 
   // Type of the object payload content
-  ObjectType object_type = 7;
+  ObjectType object_type = 7  [json_name = "Type"];
```
alexvanin, line 74: Two spaces between 7 and [

object/types.proto
```diff
@@ -93,50 +93,50 @@ message Header {
...
   // the original one is in the `Split` headers. Parent and children objects
   // must be within the same container.
   message Split {
     // Identifier of the origin object. Known only to the minor child.
-    neo.fs.v2.refs.ObjectID parent = 1;
+    neo.fs.v2.refs.ObjectID parent = 1 [json_name = "Parent"];
 
     // Identifier of the left split neighbor
-    neo.fs.v2.refs.ObjectID previous = 2;
+    neo.fs.v2.refs.ObjectID previous = 2[json_name = "Previous"];
```
alexvanin, line 112: No spaces between 2 and [

object/types.proto
```diff
@@ -93,50 +93,50 @@ message Header {
...
 
     // `signature` field of the parent object. Used to reconstruct parent.
-    neo.fs.v2.refs.Signature parent_signature = 3;
+    neo.fs.v2.refs.Signature parent_signature = 3[json_name = "ParentSignature"];
```
alexvanin, line 115: No spaces between 3 and [

object/types.proto
```diff
@@ -32,53 +32,53 @@ enum MatchType {
...
 // Object Header
 message Header {
   // Object format version. Effectively the version of API library used to
   // create particular object
-  neo.fs.v2.refs.Version version = 1;
+  neo.fs.v2.refs.Version version = 1 [json_name = "Version"];
 
   // Object's container
-  neo.fs.v2.refs.ContainerID container_id = 2;
+  neo.fs.v2.refs.ContainerID container_id = 2 [json_name = "Container"];
```
alexvanin, line 58: So it'd be like this?
```json
{
    "Container": {
        "Value": "Q29udGFpbmVySUQ="
    },
}
```

Maybe `ContainerID` name will be better there.

alexvanin (approved):
LGTM, but see the comments.

realloc:
To avoid confusion, I suggest to use the same filed naming as in protobuf definition. For example if the field is named `OwnerID` in protobuf, it should also be named `ownerID` in JSON.

cthulhu-rider (approved):

alexvanin (approved):

im-kulikov (approved):

PR #77, opened 2020-10-20
Add more well-known attributes

realloc:
Resolves nspcc-dev/neofs-api-go#172

### DIFF
--- a/container/types.proto
+++ b/container/types.proto
@@ -33,9 +33,16 @@ message Container {
   //
   // There are some "well-known" attributes affecting system behaviour:
   //
-  // * Subnet \
+  // * __NEOFS__SUBNET \
   //   String ID of container's storage subnet. Container can be attached to
   //   only one subnet.
+  //
+  // And some well-known attributes used by applications only:
+  //
+  // * Name \
+  //   Human-friendly name
+  // * Timestamp \
+  //   User-defined local time of container creation in Unix Timestamp format
   message Attribute {
     // Attribute name key
     string key = 1 [json_name = "key"];
--- a/object/types.proto
+++ b/object/types.proto
@@ -87,8 +87,19 @@ message Header {
   // There are some "well-known" attributes starting with `__NEOFS__` prefix
   // that affect system behaviour:
   //
-  // * __NEOFS__UPLOAD_ID
-  // * __NEOFS__EXPIRATION_EPOCH
+  // * __NEOFS__UPLOAD_ID \
+  //   Marks smaller parts of a split bigger object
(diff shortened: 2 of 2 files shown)


object/types.proto
```diff
@@ -87,8 +87,19 @@ message Header {
...
+  //   Marks smaller parts of a split bigger object
+  // * __NEOFS__EXPIRATION_EPOCH \
+  //   Tells GC to delete object after that epoch
+  //
+  // And some well-known attributes used by applications only:
+  //
+  // * Name \
+  //   Human-friendly name
+  // * FileName \
+  //   File name to be associated with the object on saving
```
alexvanin, line 100: Not sure about having both `Name` and `FileName`. What's the difference then?

container/types.proto
```diff
@@ -33,9 +33,16 @@ message Container {
...
-  // * Subnet \
+  // * __NEOFS__SUBNET \
   //   String ID of container's storage subnet. Container can be attached to
   //   only one subnet.
+  //
+  // And some well-known attributes used by applications only:
+  //
+  // * Name \
+  //   Human-friendly name
+  // * LocalCreationUnixTime \
```
pkorotkov, line 44: Why local in `LocalCreationUnixTime`? I believe UTC timestamps are more appropriate for that.

object/types.proto
alexvanin: File can be split for multiple objects and each split object can have some unique custom text identifier independent from original file name. Got it.

container/types.proto
realloc: Regular UTC timestamp may be encoded in lots of different formats (ISO-8601, RFC 1123, etc) while UnixTime (still in UTC) in string presentation can be encoded consistently.

pkorotkov: Well, the question is whether the value is in local time or UTC. If this is in UTC, let's put the matter to rest.

realloc: I agree that the name is misleading. It should reflect that this timestamp is created on user side without verification and may have any timestamp not related to reality.

Does anybody have suggestions on how to name it in a better way?

pkorotkov: I'd suggest not to bother with super-unombigous naming here as we simply cannot control it. A user can technically use our "well-known" attributes as they please, e.g. may write string `My birthday` to `LocalCreationUnixTime` and we must expect it to happen. Names of attributes are indicative and this one might be just `Timestamp`, for instance.

alexvanin (dismissed):

pkorotkov (approved):

PR #78, opened 2020-10-21
Clarify object search and ACL filters header references

realloc:
Describing a way to access not only object attributes, but also specific header fields.

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -295,15 +295,52 @@ message SearchRequest {
 
     // Version of the Query Language used
     uint32 version = 2;
-    // Filter structure
+    // Filter structure checks if object header field or attribute content
+    // matches a value.
+    //
+    // By default `name` field refers to the corresponding object's `Attribute`.
+    // Some Object's header fields can also be accessed by adding `$Object:`
+    // prefix to the name. Here is the list of fields available via this prefix:
+    //
+    // * $Object:version \
+    //   version
+    // * $Object:containerID \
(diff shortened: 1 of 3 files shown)


object/service.proto
```diff
@@ -295,15 +295,48 @@ message SearchRequest {
...
+    // * $Object:payloadLength \
+    //   payload_length
+    // * $Object:payloadHash \
+    //   payload_hash
+    // * $Object:objectType \
+    //   object_type
+    // * $Object:homomorphicHash \
+    //   homomorphic_hash
+    //
+    // There are some well-known filter aliases to match objects by certain propeties:
```
cthulhu-rider, line 322: I think it would be useful to additionally describe the way of specifying filter values for this class.

alexvanin: To be more specific. Now it implemented this way:
- `$Object:ROOT` key with `true` value returns only root objects, with any other value returns non root object
- `$Object:LEAF` key with `true` value returns only leaf objects, with any other value returns non leaf object
- `$Object:PARENT` key with **any** value returns objects with children list in split header (zero-object?)
- `$Object:CHILDFREE` with **any** value returns objects without children list in split header (all non-zero objects). 

Maybe we should merge last two to `CHILDFREE` with `true` and any other value.

realloc: Agree. Will fix.

alexvanin (approved):

PR #89, opened 2020-11-09
Clarify $Object:ROOT filter description

realloc:
It's expected by app developers that this filter will return the list of objects created by user, without supplementary objects like split object parts, tombstones and storage groups.

ref: nspcc-dev/neofs-node#151

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -325,11 +325,17 @@ message SearchRequest {
     // properties:
     //
     // * $Object:ROOT \
-    //   With the `value` set to `true` checks if an object is a top object in a
-    //   split hierarchy. With other values returns non-root objects.
-    // * $Object:LEAF \
-    //   With the `value` set to `true` checks if an object is a leaf in a split
-    //   hierarchy. With other values returns non-leaf objects.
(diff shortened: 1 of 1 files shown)


alexvanin (approved):

object/service.proto
```diff
@@ -326,10 +326,14 @@ message SearchRequest {
...
-    //   split hierarchy. With other values returns non-root objects.
+    //   split hierarchy. Works only with `REGULAR` type objects, including
+    //   virtual split objects. With other values returns non-root objects of
+    //   `REGULAR` type.
     // * $Object:LEAF \
     //   With the `value` set to `true` checks if an object is a leaf in a split
-    //   hierarchy. With other values returns non-leaf objects.
+    //   hierarchy. Works with all types of objects but only physically stored
+    //   on the node, without virtual split objects. With other values returns
+    //   non-leaf objects.
```
alexvanin, line 336: `LEAF` filter is not related to split hierarchy at all. 

With the `value` set to `true` returns objects that physically stored  on the node.
With other value returns virtual objects.

alexvanin (changes requested):

alexvanin (approved):

Issue #87, opened 2020-10-29
Add object ID key in extended ACL filters


alexvanin:
Extended ACL description lacks `objectID` key in filters. While it makes no sense to use such filter in `object.Search` queries, it is kinda useful in extended ACL to grant or permit access to specific object.

I think it should be `$Object:objectID` according to JSON field name:
```
message Object {
  // Object's unique identifier.
  neo.fs.v2.refs.ObjectID object_id = 1 [json_name = "objectID"];
```

PR #90, opened 2020-11-09
Add object ID key in extended ACL filters

realloc:
Extended ACL description lacks `objectID` key in filters. It's useful to control access to specific object.

Closes: #87

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -304,6 +304,8 @@ message SearchRequest {
     //
     // * $Object:version \
     //   version
+    // * $Object:objectID                       \
+    //   object_id
     // * $Object:containerID \
     //   container_id
     // * $Object:ownerID \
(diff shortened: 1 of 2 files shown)


object/service.proto
```diff
@@ -304,6 +304,8 @@ message SearchRequest {
     //
     // * $Object:version \
     //   version
+    // * $Object:objectID                       \
+    //   object_id
```
alexvanin, line 308: Are there any cases when it is useful to make search request with `$Object:objectID` filter when you can use less exhaustive `object.Head` request?

realloc: Search for objects in a group except the known set. For example search for other object parts.

alexvanin: `$Object:objectID StringNotEqual <some-object-id>` makes sense. We don't have `StringNotEqual` matcher yet, but if it is planned, then I don't mind having `$Object:objectID ` already in the list.

alexvanin (approved):

PR #102, opened 2020-12-09
Remove ACL filter unusable in ACL context

### LINKED ISSUE #85: Leave only root object related fields for ACL filter
ACL filter defines these keys to filter objects (outside of user-defiend headers):
- version
- container ID
- owner ID
- creation epoch
- payload length
- payload hash
- object type
- homomorphic hash

It mostly copies object search filters, however not all of them are quite applicable for ACL tables. Say there is no point to make ACL rules based on hashes or object type (ACL rules seems legit for regular objects only). Maybe we should short this list?

realloc:
Removing `$Object:containerID`, because requests are already limited to container scope and NeoFS doesn't support cross-container operations.

Closes #85

### DIFF
--- a/acl/types.proto
+++ b/acl/types.proto
@@ -106,8 +106,6 @@ message EACLRecord {
   //   version
   // * $Object:objectID \
   //   object_id
-  // * $Object:containerID \
-  //   container_id
   // * $Object:ownerID \
   //   owner_id
   // * $Object:creationEpoch \


alexvanin:
I would like to freeze this change for a while to check if it is okay to remove such filter. Maybe it will require more changes in node than expected.

realloc:
It's safe to leave `ContainerID` filter and keep backward compatibility.  Even if it's useless, it's still harmless.

Issue #109, opened 2020-12-16
Define audit result structure in API


alexvanin:
Inner ring nodes make data audit and produce audit results. These results are not transferred between nodes but saved in audit contract. Later settlement routines will transfer payments based on these results.

Audit result structure should contain:
- epoch number (consider fixed size integer),
- container ID,
- public key of inner ring node that produced audit result,
- list of storage group object IDs that **passed** PoR,
- list of storage group object IDs that **failed** PoR,
- amount of hits at PoP,
- amount of misses at PoP,
- amount of fails at PoP,
- list of storage node public keys that **passed at least one** PDP,
- list of storage node public keys that **failed at least one** PDP.

PR #114, opened 2020-12-17
Define data audit result structure

realloc:
Inner ring nodes conduct data audit and submit audit results. Results are then saved in audit smart contract for settlement routines to transfer payments.

We don't need to save the full detailed audit report, but only provide enough information for other subsystems to reward or punish storage nodes.

Closes: #109

### DIFF
--- a/audit/types.proto
+++ b/audit/types.proto
@@ -1,0 +1,46 @@
+syntax = "proto3";
+
+package neo.fs.v2.audit;
+
+option go_package = "github.com/nspcc-dev/neofs-api-go/v2/audit/grpc;audit";
+option csharp_namespace = "NeoFS.API.v2.Audit";
+
+import "refs/types.proto";
+
+// DataAuditResult keeps record of conducted Data Audits. The detailed report is
+// generated separately.
+message DataAuditResult {
+  // Epoch number when the Data Audit was conducted
+  fixed64 audit_epoch = 1 [json_name = "auditEpoch"];
+
+  // Container under audit
+  neo.fs.v2.refs.ContainerID container_id = 2 [json_name = "containerID"];
+
+  // Public key of the auditing InnerRing node in a binary format
+  bytes public_key = 3 [json_name = "publicKey"];
+
+  // List of Storage Groups that passed audit PoR stage
+  repeated neo.fs.v2.refs.ObjectID pass_sg = 4 [json_name = "passSG"];
+
+  // List of Storage Groups that failed audit PoR stage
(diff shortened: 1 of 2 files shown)


audit/types.proto
```diff
@@ -0,0 +1,46 @@
...
+option go_package = "github.com/nspcc-dev/neofs-api-go/v2/audit/grpc;audit";
+option csharp_namespace = "NeoFS.API.v2.Audit";
+
+import "refs/types.proto";
+
+// DataAuditResult keeps record of conducted Data Audits. The detailed report is
+// generated separately.
+message DataAuditResult {
+  // Epoch number when the Data Audit was conducted
+  uint64 audit_epoch = 1 [json_name = "auditEpoch"];
```
alexvanin, line 14: To avoid varint issues in contract I propose to use `fixed64` type for epochs.

audit/types.proto
```diff
@@ -0,0 +1,46 @@
...
+  uint32 miss = 7 [json_name = "miss"];
+
+  // Number of objects under audit stored in a way not confirming placement
+  // policy or not found at all when checking PoP
+  uint32 fail = 8 [json_name = "fail"];
+
+  // List of storage node public keys that passed at least one PDP
+  repeated bytes pass_nodes = 9 [json_name = "passNodes"];
+
+  // List of storage node public keys that passed at least one PDP
```
alexvanin, line 44: that **failed** at least one PDP

audit/types.proto
```diff
@@ -0,0 +1,46 @@
...
+  // containers placement policy when checking PoP
+  uint32 hit = 6 [json_name = "hit"];
+
+  // Number of objects under audit placed in suboptimal way according to the
+  // containers placement policy, but still at a satisfactory level when
+  // checking PoP
+  uint32 miss = 7 [json_name = "miss"];
+
+  // Number of objects under audit stored in a way not confirming placement
+  // policy or not found at all when checking PoP
```
alexvanin, line 38: Maybe we should notice that `hit`, `miss` and `fail` are counters in audit **sample** and it is not precise values? E.g. `Number of objects under audit sample stored ...`

alexvanin (changes requested):

audit/types.proto
realloc: fixed

alexvanin (approved):

cthulhu-rider (dismissed):

alexvanin (approved):

PR #118, opened 2021-01-11
Describe well-known X-Headers

### LINKED ISSUE #117: Describe well-known x-headers for request epoch number
To implement https://github.com/nspcc-dev/neofs-node/issues/234 there should be two well-known x-header keys with prefix.
These x-headers represent **epoch of netmap to use for lookup** and **the depth of previous epoch lookup** (storage node looks for objects in nodes of previous epoch container if it can't find it anywhere). 

Define string representation of these x-header values. 

Mention, that it is useful first of all for read operations.

realloc:
When object can't be found using current network map, there is a chance to find it using netmap versions from previous epochs.

By default only current netmap is used. Well-known X-Headers can be used to set the specific netmap version and limit how deep into the past the node can go in the search of an object.

Closes #117

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -85,7 +85,25 @@ message SessionToken {
   neo.fs.v2.refs.Signature signature = 2 [json_name = "signature"];
 }
 
-// Extended headers for Request/Response.
+// Extended headers for Request/Response. May contain any user-defined headers
+// to be interpreted on application level.
+//
+// Key name must be unique valid UTF-8 string. Value can't be empty. Requests or
+// Responses with duplicated header names or headers with empty values will be
+// considered invalid.
+//
+// There are some "well-known" headers starting with `__NEOFS__` prefix that
+// affect system behaviour:
+//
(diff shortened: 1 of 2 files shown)


session/types.proto
```diff
@@ -85,7 +85,23 @@ message SessionToken {
...
+//
+// Key name must be unique valid UTF-8 string. Value can't be empty. Requests or
+// Responses with duplicated header names or headers with empty values will be
+// considered invalid.
+//
+// There are some "well-known" headers starting with `__NEOFS__` prefix that
+// affect system behaviour:
+//
+// * __NEOFS__NETMAP_EPOCH \
+//   Netmap epoch to use for object placement calculation. If set to '0' or not
```
alexvanin, line 99: It would be nice to specify format of the string value for both `NETMAP_EPOCH` and `NETMAP_LOOKUP_DEPTH`. E.g. add `Value represented as decimal number` lines.

cthulhu-rider (approved):

PR #126, opened 2021-02-02
object: Define NOT_PRESENT value in MatchType enum

### LINKED ISSUE #120: Define NOT_PRESENT match type of object search filter
`NOT_PRESENT` match type ignores values and looks if object has specified key.

```
Filter: A NOT_PRESENT 

Object1(not matched)    Object2(not matched)     Object3(matched)   
A: One                  A:                       B: Two
B: Two                  B: Two                   C: Three
C: Three                C: Three                 D: Four
```

cthulhu-rider:
Closes #120.

### DIFF
--- a/object/types.proto
+++ b/object/types.proto
@@ -38,6 +38,9 @@ enum MatchType {
 
   // Full string mismatch
   STRING_NOT_EQUAL = 2;
+
+  // Lack of key
+  NOT_PRESENT = 3;
 }
 
 // Short header fields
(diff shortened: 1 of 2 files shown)


object/types.proto
```diff
@@ -38,6 +38,9 @@ enum MatchType {
 
   // Full string mismatch
   STRING_NOT_EQUAL = 2;
+
+  // Lack of value
```
alexvanin, line 42: It is lack of _key_, isn't it?

cthulhu-rider: Yes, it is, fixed

alexvanin (approved):

PR #140, opened 2021-03-10
change csharp namespace

ZhangTao1596:
Signed-off-by: ZhangTao1596 <zhangtao@ngd.neo.org>

Chane csharp namespace to **Neo.FileSystem.API**

### DIFF
--- a/accounting/service.proto
+++ b/accounting/service.proto
@@ -3,7 +3,7 @@ syntax = "proto3";
 package neo.fs.v2.accounting;
 
 option go_package = "github.com/nspcc-dev/neofs-api-go/v2/accounting/grpc;accounting";
-option csharp_namespace = "NeoFS.API.v2.Accounting";
+option csharp_namespace = "Neo.FileSystem.API.Accounting";
 
 import "accounting/types.proto";
 import "refs/types.proto";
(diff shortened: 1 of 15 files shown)


realloc (changes requested):
NeoFS is not a File System. It must not be named FileSystem to avoid confusion. Please change the naming.

realloc:
Also, is it possible to consider using API version in the package? Is it feasible for C#? There may be other API versions in future.

ZhangTao1596:
> NeoFS is not a File System. It must not be named FileSystem to avoid confusion. Please change the naming.

Isn't `FS` for `FileSystem`?  What name do you suggest?

ZhangTao1596:
> I'd suggest to use NeoFS is it is, without decoding the abbreviation.
> 
> According to NeoFS Specification, NeoFS stands for NeoFS Exaggerative Object File Storage, also known as Neo File Storage :)

What about `Neo.FileStorage`. Because all neo official packages on `nuget.org`start with `Neo.`.

realloc:
> What about `Neo.FileStorage`. Because all neo official packages on `nuget.org`start with `Neo.`.

Sounds good, but, maybe just, `Neo.FS.API.*` would be shorter =) I'm OK with both options.

realloc (approved):

PR #158, opened 2021-05-24
Support container sessions

### LINKED ISSUE #156: Define container context in session token
Container operations are similar to object operations in terms of permission delegation. Protocol gate implementations require permission delegation to operate with data on behalf of user. 

Permission delegation is done by session token. We should expand session token body with container context. Container context includes:
- verb (Put, Delete, SetEACL),
- container ID.

We ignore all reading verbs because data publicly available on side chain, therefore it is useless to grant access for such operations. 

There is a special container ID `JEKNVnkbo3jma5nREBBJCDoXFVeKkD56V3xKrvRmWxFG` (32 bytes of 0xFF) that should be interpreted as a wild card for the node. Node with delegated permissions can do specified operation with any user's container.

### LINKED ISSUE #157: Provide container session token to and from container service
There are several ways container session token can be provided to and by container service.

## Include session token in container and EACL bodies.

Objects already has session token field in the header, so it makes sense to do the same way with containers and EACL. This way we won't change signatures of RPC services and contract methods.

## Include session token in meta headers.

The other way to include session tokens in request and response meta headers. In this case we should update container contract signatures and make additional storage for session tokens there. Also responses do not have token field in meta header yet. It actually doesn't make much sense, because it is meta information and it is not represented in response structures. It is not the case for the session token.

---

Personally I would go with option (1), because it requires small changes (two extra fields) and it is robust.

cthulhu-rider:
Closes #156 
Closes #157.

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -102,6 +102,9 @@ message SessionToken {
     oneof context {
       // ObjectService session context
       ObjectSessionContext object = 5 [json_name = "object"];
+
+      // ContainerService session context
+      ContainerSessionContext container = 6 [json_name = "container"];
     }
   }
   // Session Token contains the proof of trust between peers to be attached in
--- a/session/types.proto
+++ b/session/types.proto
@@ -43,6 +43,34 @@ message ObjectSessionContext {
   neo.fs.v2.refs.Address address = 2 [json_name = "address"];
 }
 
+// Context information for Session Tokens related to ContainerService requests.
+message ContainerSessionContext {
+    // Container request verbs
+    enum Verb {
+        // Unknown verb
+        VERB_UNSPECIFIED = 0;
+
+        // Refers to container.Put RPC call
+        PUT = 1;
+
+        // Refers to container.Delete RPC call
+        DELETE = 2;
+
+        // Refers to container.SetExtendedACL RPC call
+        SETEACL = 3;
+    }
+    // Type of request for which the token is issued
+    Verb verb = 1 [json_name = "verb"];
+
+    // Spreads the action to all owner containers.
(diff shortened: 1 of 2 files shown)


session/types.proto
```diff
@@ -43,6 +43,36 @@ message ObjectSessionContext {
   neo.fs.v2.refs.Address address = 2 [json_name = "address"];
 }
 
+// Context information for Session Tokens related to ContainerService requests
```
realloc, line 46: There should be a dot at the end of the sentence

session/types.proto
```diff
@@ -43,6 +43,36 @@ message ObjectSessionContext {
...
+        // Refers to container.Delete RPC call
+        DELETE = 2;
+
+        // Refers to container.SetExtendedACL RPC call
+        SETEACL = 3;
+    }
+    // Type of request for which the token is issued
+    Verb verb = 1 [json_name = "verb"];
+
+    // Spreads the action to all owner containers.
```
realloc, line 65: Breaks pandoc format. Must be one line.

session/types.proto
```diff
@@ -43,6 +43,36 @@ message ObjectSessionContext {
...
+    //
+    // If set, container_id field is ignored.
+    bool wildcard = 2 [json_name = "wildcard"];
+
+    // Particular container to which the action applies.
+    //
```
realloc, line 71: Breaks pandoc format. Must be one line.

realloc (changes requested):
LGTM, but requiers minor documentation comments format changes

alexvanin (approved):

realloc (approved):

PR #174, opened 2021-09-15
Expand netmap.NetworkInfo

### LINKED ISSUE #172: Expand NetworkInfo information with more data
After https://github.com/nspcc-dev/neo-go/issues/2160 we can expand `NetworkInfo` message with details about side chain:
- Ms per block,
- ... ?

Also we can include NeoFS global configuration records as a list of key/value (both repeated bytes).
Keys are hex encoded strings and values ara raw data from netmap contract.

cthulhu-rider:
Closes #172.

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -235,4 +235,7 @@ message NetworkInfo {
 
     // Magic number of the sidechain of the NeoFS network.
     uint64 magic_number = 2 [json_name = "magicNumber"];
+
+    // MillisecondsPerBlock network parameter of the sidechain of the NeoFS network.
+    int64 ms_per_block = 3 [json_name = "msPerBlock"];
 }
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -252,4 +252,7 @@ message NetworkInfo {
 
     // MillisecondsPerBlock network parameter of the sidechain of the NeoFS network.
     int64 ms_per_block = 3 [json_name = "msPerBlock"];
+
+    // NeoFS network configuration.
+    NetworkConfig network_config = 4 [json_name = "networkConfig"];
 }
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -228,6 +228,20 @@ message NodeInfo {
   State state = 4 [json_name = "state"];
 }
 
+// NeoFS network configuration
+message NetworkConfig {
+  // Single configuration parameter.
+  message Parameter {
+    // Parameter key. UTF-8 encoded string.
+    bytes key = 1 [json_name = "key"];
+
+    // Parameter value.
+    bytes value = 2 [json_name = "value"];
+  }
+  // List of parameter values.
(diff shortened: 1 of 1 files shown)


netmap/types.proto
```diff
@@ -228,11 +228,31 @@ message NodeInfo {
...
 // Information about NeoFS network
 message NetworkInfo {
     // Number of the current epoch in the NeoFS network.
     uint64 current_epoch = 1 [json_name = "currentEpoch"];
 
     // Magic number of the sidechain of the NeoFS network.
     uint64 magic_number = 2 [json_name = "magicNumber"];
+
+    // MillisecondsPerBlock network parameter.
+    int64 ms_per_block = 3 [json_name = "msPerBlock"];
```
realloc, line 254: Maybe it would be better to have the Epoch duration in blocks here and get ms per block from chain parameters in NetworkConfig?

cthulhu-rider: I'm not sure what structure is better. I definitely want to separate the standard network parameters and the NeoFS-specific ones, and which of them will be in the `NetworkConfig` is under discussion. Current structure represents issue proposal. @alexvanin  what do u think?

alexvanin: > I definitely want to separate the standard network parameters and the NeoFS-specific ones

Agree with that. As soon as the Epoch duration in blocks is a part of NeoFS Global configuration, I would not expect to see it in a separate field.

alexvanin: I guess there is a bit of misunderstanding in the naming. `NetworkConfig` is a Global NeoFS Configuration, while other fields related to the chain (except current epoch).

netmap/types.proto
```diff
@@ -228,11 +228,31 @@ message NodeInfo {
...
+
 // Information about NeoFS network
 message NetworkInfo {
     // Number of the current epoch in the NeoFS network.
     uint64 current_epoch = 1 [json_name = "currentEpoch"];
 
     // Magic number of the sidechain of the NeoFS network.
     uint64 magic_number = 2 [json_name = "magicNumber"];
+
+    // MillisecondsPerBlock network parameter.
```
alexvanin, line 253: I suggest to emphasize that it is side chain network parameter in the comment. Therefore `network_config` is a NeoFS network configuration.

netmap/types.proto
```diff
@@ -228,11 +228,31 @@ message NodeInfo {
   State state = 4 [json_name = "state"];
 }
 
+// NeoFS network configuration
+message NetworkConfig {
+  // Single configuration parameter.
+  message Parameter {
+    // Parameter key. Hex-encoded string.
```
alexvanin, line 235:
```suggestion
    // Parameter key. UTF-8 encoded string.
```

cthulhu-rider: Fixed

alexvanin (approved):

realloc (approved):

PR #175, opened 2021-09-27
session: Add sidechain magic to RequestMetaHeader

### LINKED ISSUE #82: Sidechain Magic number should be used in NeoFS requests
To prevent theoretical replay attack, we need to have Sidechain Magic in NeoFS requests.

cthulhu-rider:
Closes #82.

Please note that a new header field has been added before the origin field, although it has a higher ordinal number. In my opinion, this order will make it easier to understand the structure of the header.

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -166,6 +166,9 @@ message RequestMetaHeader {
 
   // `RequestMetaHeader` of the origin request
   RequestMetaHeader origin = 7 [json_name = "origin"];
+
+  // Network magic of NeoFS sidechain
+  uint64 sidechain_magic = 8 [json_name = "sidechainMagic"];
 }
 
 // Information about the response


alexvanin:
> Please note that a new header field has been added before the origin field, although it has a higher ordinal number. In my opinion, this order will make it easier to understand the structure of the header.

While it may be easier to understand the structure, mixed field order looks quite confusing. I think the only right solution to keep the order and the structure was to do this:

```
  neo.fs.v2.acl.BearerToken bearer_token = 6;

  uint64 sidechain_magic = 7;

  RequestMetaHeader origin = 99;
```

However we deliberately decided to not to do this for some reason, therefore we should live with jumbled fields.

alexvanin (approved):

PR #182, opened 2021-11-17
netmap: add subnet id to placement policy

### LINKED ISSUE #179: Expand placement policy with subnetwork field
The best way to filter storage nodes of subnetwork is to define subnetwork ID directly in placement policy. 

UPD 16.11 @cthulhu-rider

1. Define `refs.SubnetID` type with single numerical field and describe:
- string format;
- binary format.

Null value should refer to zero subnet.

2. Add `subnet` field to `PlacementPolicy` message. Write the docs how it affects placement building.

fyrchik:
Close #179 .

### DIFF
--- a/refs/types.proto
+++ b/refs/types.proto
@@ -77,6 +77,16 @@ message OwnerID {
   bytes value = 1 [json_name = "value"];
 }
 
+// NeoFS subnetwork identifier.
+//
+// String representation of a value is base-10 integer.
+//
+// JSON representation is an object containing single `value` number field.
+message SubnetID {
+  // 4-byte integer subnetwork identifier.
+  fixed32 value = 1 [json_name = "value"];
+}
+
 // API version used by a node.
 //
 // String presentation is a Semantic Versioning 2.0.0 compatible version string
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -5,6 +5,8 @@ package neo.fs.v2.netmap;
 option go_package = "github.com/nspcc-dev/neofs-api-go/v2/netmap/grpc;netmap";
 option csharp_namespace = "Neo.FileStorage.API.Netmap";
 
+import "refs/types.proto";
+
 // Operations on filters
 enum Operation {
   // No Operation defined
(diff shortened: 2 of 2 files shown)


realloc (approved):

refs/types.proto
```diff
@@ -77,6 +77,18 @@ message OwnerID {
   bytes value = 1 [json_name = "value"];
 }
 
+// NeoFS subnetwork identifier.
+//
+// `SubnetID` is a 4-byte long.
+//
+// String representation is base-10 integer value.
+//
+// JSON representation is a JSON number.
```
cthulhu-rider, line 86: Suffice it to say that it corresponds to the standard Protocol Buffers JSON format. If you want to explicitly describe the format of this message, then this is an object with a numeric field `value`.

refs/types.proto
```diff
@@ -77,6 +77,18 @@ message OwnerID {
   bytes value = 1 [json_name = "value"];
 }
 
+// NeoFS subnetwork identifier.
+//
+// `SubnetID` is a 4-byte long.
```
cthulhu-rider, line 82: Suffice it to say that it corresponds to the standard NeoFS binary format which is Protocol Buffers with direct orders of fields (I'm not sure it is actually 4-byte long because of prefixes).

refs/types.proto
```diff
@@ -77,6 +77,18 @@ message OwnerID {
...
+//
+// String representation is base-10 integer value.
+//
+// JSON representation is a JSON number.
+message SubnetID {
+  // Integer subnetwork identifier.
+  fixed32 value = 1 [json_name = "value"];
```
cthulhu-rider, line 89: Let's clearly indicate that zero identifies the zero subnet.

alexvanin (approved):

refs/types.proto
fyrchik: This is more about the `PlacementPolicy` interpretatiion than the id itself.

refs/types.proto
fyrchik: Doc shouldn't really describe the format (we are talking about `.proto` file after all), but having explicit id size is something I would like to know from the doc. Moved this to the field comment.

refs/types.proto
alexvanin: > why have you chosen `fixed32`?

`fixed32` was mentioned for in-place ID checks in smart contract. With `uint32` type ID will be encoded as varint, which is much harder to decode.

PR #186, opened 2022-01-10
Pass sidechain magic in requests

### LINKED ISSUE #82: Sidechain Magic number should be used in NeoFS requests

cthulhu-rider:
* closes #82

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -167,6 +167,10 @@ message RequestMetaHeader {
 
   // `RequestMetaHeader` of the origin request
   RequestMetaHeader origin = 7 [json_name = "origin"];
+
+  // NeoFS network magic. Must match the value for the network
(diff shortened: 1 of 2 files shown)


session/types.proto
```diff
@@ -167,6 +167,10 @@ message RequestMetaHeader {
 
   // `RequestMetaHeader` of the origin request
   RequestMetaHeader origin = 7 [json_name = "origin"];
+
+  // Network magic of NeoFS sidechain. Must match the value for the network
+  // that the server belongs to.
+  uint64 sidechain_magic = 8 [json_name = "sidechainMagic"];
```
realloc, line 173: Maybe `network_magic` or just `magic` would be better?

Issue #191, opened 2022-02-07
Create status codes for session-related errors


fyrchik:
Currently there are 2 candidates https://github.com/nspcc-dev/neofs-node/blob/master/pkg/services/object/util/key.go#L13

alexvanin:
Also `errSessionTokenExpired` may be redundant on API level. Storage node will never produce such error after https://github.com/nspcc-dev/neofs-node/issues/1133. If token is expired but still available on the storage node, it can return API error of token not found anyway.

fyrchik:
I think there is still a race condition in case of token expiration, so it is possible to get expired token from storage.

alexvanin:
> I think there is still a race condition in case of token expiration, so it is possible to get expired token from storage.

It is. My point was that those errors are pretty much the same? If it is expired and if it was not found, they both indicate that client has to renew session token. Maybe it can be encoded as a detail, not the separate error code. I am not completely sure about that, though.

fyrchik:
I agree they are essentially the same, just that an error message in case of expired token can be confusing. Maybe user-facing software should just tell `token was not found  or has expired`, this way one status is enough. Anyway both errors should be processed in an equal fashion.

Issue #190, opened 2022-01-27
Status code for not found


KirillovDenis:
It would be nice to have status code for `not found` container and object errors as well as for #189.

alexvanin:
Also `ErrAlreadyRemoved`, see https://github.com/nspcc-dev/neofs-node/issues/1158

PR #192, opened 2022-02-09
Add status codes for object, container and session services.

### LINKED ISSUE #189: Status code for ACL access restrcitction
I guess we can have one or two (for basic and extended ACL) status codes that notify about access restriction. It can be used in https://github.com/nspcc-dev/neofs-s3-gw/pull/328 to return correct status code (now it relies on message with `access to operation * is denied by` sub string).

fyrchik:
Close #189, #190, #191.

1. I am not sure whether ACL-related codes should be in the Object, Container or a separate ACL section. `object.Put` acl denied seems to belong to container service, while `object.Get` is related to a specific object.
2. Prefixes are somewhat ugly IMO but necessary because of name conflicts. May be we should prefix all enum values with the service name.
3. We can also define service specific errors in related packages, which helps us with (2). Here I define them in one package.

### DIFF
--- a/status/types.proto
+++ b/status/types.proto
@@ -67,6 +67,9 @@ enum Section {
 
     // Object service-specific errors.
     SECTION_OBJECT = 2;
+
+    // Container service-specific errors.
+    SECTION_CONTAINER = 3;
 }
 
 // Section of NeoFS successful return codes.
@@ -100,4 +103,9 @@ enum Object {
     // [**2049**] Object not found.
     OBJECT_NOT_FOUND = 1;
 }
+
+// Section of statuses for container-related operations.
+enum Container {
+    // [**3072**] Container not found.
+    CONTAINER_NOT_FOUND = 0;
 }
--- a/status/types.proto
+++ b/status/types.proto
@@ -64,6 +64,9 @@ enum Section {
 
     // Failure codes regardless of the operation.
     SECTION_FAILURE_COMMON = 1;
+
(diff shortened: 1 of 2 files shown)


cthulhu-rider:
1. I suggest to start with localizing access control codes in the object section.
3. I'd place service-specific codes in corresponding packages, and place generic ones in `status` package.

cthulhu-rider:
Lets add docs to related RPCs.

status/types.proto
```diff
@@ -87,3 +96,27 @@ enum CommonFail {
...
+    CONTAINER_NOT_FOUND = 0;
+}
+
+// Section of statuses for session-related operations.
+enum Session {
+    // [**4096**] Token not found.
+    TOKEN_NOT_FOUND = 0;
+    // [**4097**] Token has expired.
+    TOKEN_EXPIRED = 1;
+}
```
cthulhu-rider, line 122: \n

cthulhu-rider (approved):

carpawell:
was it decided to not add [ErrAlreadyRemoved](https://github.com/nspcc-dev/neofs-api/issues/190#issuecomment-1034715600)?

cthulhu-rider:
> was it decided to not add [ErrAlreadyRemoved](https://github.com/nspcc-dev/neofs-api/issues/190#issuecomment-1034715600)?

Missed this one, will be added in next PR.

Issue #193, opened 2022-02-11
Define general behavior for object notifications


alexvanin:
NeoFS Storage node should support notification extension: optional notification channel to produce events. This channel and event structure defined depending on extension implementation. API should define general behavior for object notifications (maybe in neofs-spec?) and well-known object attributes.

General behavior: 
- Notification event consists of topic and message
- There is only object notifications for now
- Object notification message is always object address (format defined by notification extension)
- Object notification topic is defined in `__NEOFS__TICK_TOPIC` header (UTF-8 string)
- Node produces object notifications at epoch defined in `__NEOFS__TICK_EPOCH` header which is Base10 encoded number
- Node produces object notification if both `__NEOFS__TICK_TOPIC` and `__NEOFS__TICK_EPOCH` are set in object attributes

realloc:
If there is only `TOPIC` attribute set, it will have no effect and will never be taken into account.
If there is only `TICK` attribute set, it should send event to the default topic, as configured in Storage Node settings. The default notification settings may be different depending on the transport and are out of the scope of API description.

carpawell:
> If there is only `TICK` attribute set, it should send event to the default topic

what should a node do when there is no default topic in the config:
1. do not start at all;
2. start and try to fetch topic from attributes, if there is no such attr, skip it with a warning?

realloc:
> what should a node do when there is no default topic in the config:

That's incorrect configuration. Since each config parameter must have a default attribute (if it makes sense in each particular cease), such situation can't arise normally.

For node's default topic default value let's use it's node id.

cthulhu-rider:
Moved to neofs-spec.

PR #195, opened 2022-02-14
Feat/Add notification object attributes

carpawell:
Related to #193.

I think it should not be included in the next (`2.11.0`+) release, so it may wait.

### DIFF
--- a/object/types.proto
+++ b/object/types.proto
@@ -117,6 +117,13 @@ message Header {
   //   Marks smaller parts of a split bigger object
   // * __NEOFS__EXPIRATION_EPOCH \
   //   Tells GC to delete object after that epoch
+  // * __NEOFS__TICK_EPOCH \
+  //   Decimal number that defines what epoch must produce
+  //   object notification with UTF-8 object address in a
+  //   body (`0` value produces notification right after
+  //   object put)
+  // * __NEOFS__TICK_TOPIC \
+  //   UTF-8 string topic ID that is used for object notification
   //
   // And some well-known attributes used by applications only:
   //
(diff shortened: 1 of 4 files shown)


object/types.proto
```diff
@@ -117,6 +117,12 @@ message Header {
   //   Marks smaller parts of a split bigger object
   // * __NEOFS__EXPIRATION_EPOCH \
   //   Tells GC to delete object after that epoch
+  // * __NEOFS__TICK_TOPIC \
+  //   UTF-8 string topic ID that is used for object notification
+  // * __NEOFS__TICK_EPOCH \
+  //   Base10 encoded number that defines what epoch must produce
```
realloc, line 123: Base10 -> decimal

object/types.proto
```diff
@@ -117,6 +117,12 @@ message Header {
   //   Marks smaller parts of a split bigger object
   // * __NEOFS__EXPIRATION_EPOCH \
   //   Tells GC to delete object after that epoch
+  // * __NEOFS__TICK_TOPIC \
```
realloc, line 120: It' better to put it after `TICK_EPOCH` attribute, to simplify understanding.

object/types.proto
```diff
@@ -117,6 +117,12 @@ message Header {
...
+  //   UTF-8 string topic ID that is used for object notification
+  // * __NEOFS__TICK_EPOCH \
+  //   Base10 encoded number that defines what epoch must produce
+  //   object notification (`0` value produces notification right
+  //   after object put)
```
realloc, line 125: I'd also add a note that the notification body will contain object address in UFT-8 string.

realloc (changes requested):

object/types.proto
carpawell: thought such details should be in spec, but added

realloc: NeoFS Spec is automatically generated from this API definitions. =)

carpawell: i meant a few manually written words about notifications

cthulhu-rider (approved):

fyrchik (approved):

realloc (approved):

PR #196, opened 2022-02-15
object: Add `LOCK` type

### LINKED ISSUE #194: Define LOCK object type
LOCK object type is an object type that restricts object removal. Its payload should be a list of objects that should not be deleted. 
If user sends `object.Delete` request on a object which is in this list, then it should return error.

**Question:** should `object.Put(tombstone)` also search for LOCK objects? Or metabase may deny tombstones if there is a LOCK object already /cc @realloc 

If Storage node's GC wants to remove an object (due to expiration epoch, for example), it should not do that if LOCK object is present.

LOCK objects distributed over all container nodes the same way as tombstones.

cthulhu-rider:
* closes #194

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -35,7 +35,13 @@ service ObjectService {
   // Statuses:
   // - **OK** (0, SECTION_SUCCESS):
   // object has been successfully saved in the container;
-  // - Common failures (SECTION_FAILURE_COMMON).
+  // - Common failures (SECTION_FAILURE_COMMON);
+  // - **LOCKED** (2048, SECTION_OBJECT):
+  // placement of an object of type TOMBSTONE that includes at least one locked
+  // object is prohibited;
+  // - **LOCK_NON_REGULAR_OBJECT** (2049, SECTION_OBJECT):
+  // placement of an object of type LOCK that includes at least one object of
+  // type other than REGULAR is prohibited.
   rpc Put(stream PutRequest) returns (PutResponse);
 
   // Delete the object from a container. There is no immediate removal
@@ -44,7 +50,9 @@ service ObjectService {
   // Statuses:
   // - **OK** (0, SECTION_SUCCESS):
   // object has been successfully marked to be removed from the container;
-  // - Common failures (SECTION_FAILURE_COMMON).
+  // - Common failures (SECTION_FAILURE_COMMON);
+  // - **LOCKED** (2048, SECTION_OBJECT):
+  // deleting a locked object is prohibited.
   rpc Delete(DeleteRequest) returns (DeleteResponse);
 
   // Returns the object Headers without data payload. By default full header is
--- a/object/service.proto
+++ b/object/service.proto
@@ -22,7 +22,15 @@ service ObjectService {
   // Statuses:
   // - **OK** (0, SECTION_SUCCESS):
   // object has been successfully read;
-  // - Common failures (SECTION_FAILURE_COMMON).
+  // - Common failures (SECTION_FAILURE_COMMON);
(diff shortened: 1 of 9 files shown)


cthulhu-rider:
I've added documentation of `Object` payload format for `LOCK` types. After the merge I will create an issue to add docs for all other types.

object/types.proto
```diff
@@ -204,3 +216,9 @@ message SplitInfo {
   // object parts.
   neo.fs.v2.refs.ObjectID link = 3;
 }
+
+// Lock keeps record of objects that are locked.
+message Lock {
```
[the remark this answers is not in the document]
realloc: Agree, let's move it to a separate directory.

status/types.proto
```diff
@@ -87,3 +90,12 @@ enum CommonFail {
     //  unsigned integer).
     WRONG_MAGIC_NUMBER = 1;
 }
+
+// Section of failed statuses for object-related operations.
+enum Object {
+    // [**2048**] Operation rejected by the object lock.
+    LOCKED = 0;
```
realloc, line 97: Is this for locking a locked object?

status/types.proto
```diff
@@ -87,3 +90,12 @@ enum CommonFail {
...
+
+    // [**2049**] Locking of the irregular object rejected.
+    IRREGULAR_LOCK = 1;
```
realloc, line 100: `WRONG_OBJECT_TYPE`? The "irregular" may be misinterpreted.

object/types.proto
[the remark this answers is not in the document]
cthulhu-rider: May be we need to concentrate object content types in single package? Otherwise, for the sake of one type, we will create a new package/dir. Also it seems more

status/types.proto
cthulhu-rider: Then we need some details since `WRONG_OBJECT_TYPE` is too generic.

object/types.proto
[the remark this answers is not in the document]
realloc: There may be service calls added in the future, related to each type. New object types are rare, we don't expect them to grow in number in the nearest future, so let's have a new directory for locks.

status/types.proto
realloc: `WRONG_LOCKED_OBJECT_TYPE`?

cthulhu-rider: Renamed to `LOCK_NON_REGULAR_OBJECT`

realloc (approved):
Lets the others also review this.

carpawell (approved):

realloc (approved):

object/service.proto
```diff
@@ -75,7 +124,17 @@ service ObjectService {
...
+  // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER):
+  // object container not found;
+  // - **ACCESS_DENIED** (2048, SECTION_OBJECT):
+  // access to operation RANGE of the object is denied;
+  // - **OBJECT_NOT_FOUND** (2049, SECTION_OBJECT):
+  // object not found in container;
+  // - **TOKEN_EXPIRED** (4097, SECTION_SESSION):
+  // provided session token has expired;
+  // - **OBJECT_ALREADY_REMOVED** (2052, SECTION_OBJECT):
+  // the requested object has been marked as deleted.
```
fyrchik, line 137: Sort them by section and/or id?

fyrchik (approved):

object/service.proto
cthulhu-rider: By ID is useless IMO since there are "holes" in numbers. By section - agree, I'd be convenient.
https://github.com/nspcc-dev/neofs-api/issues/198

PR #197, opened 2022-02-21
[#55] refs: add signature scheme description

### LINKED ISSUE #55: Add key and signature types into Signature message
NeoFS uses different key and signature types, but this knowledge is not reflected directly in protocol definitions, only in code.

`Signature` structure needs to have key type reference and signature algorithm reference.

fyrchik:
Close #55.

Signed-off-by: Evgenii Stratonikov <evgeniy@nspcc.ru>

### DIFF
--- a/refs/types.proto
+++ b/refs/types.proto
@@ -105,6 +105,18 @@ message Signature {
   bytes key = 1 [json_name = "key"];
   // Signature
   bytes sign = 2 [json_name = "signature"];
+  // Scheme contains digital signature scheme identifier.
+  SignatureScheme scheme = 3 [json_name = "scheme"];
+}
+
+// Signature scheme describes digital signing scheme used for (key, signature) pair.
+enum SignatureScheme {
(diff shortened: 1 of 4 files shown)


refs/types.proto
```diff
@@ -105,6 +105,18 @@ message Signature {
   bytes key = 1 [json_name = "key"];
   // Signature
   bytes sign = 2 [json_name = "signature"];
+  // Scheme contains digital signature scheme identifier.
+  SignatureScheme scheme = 3 [json_name = "scheme"];
+}
+
+// Signature scheme describes digital signing scheme used for (key, signature) pair.
+enum SignatureScheme {
+  // Unknown.
```
realloc, line 114: Let's add a comment clarifying that `UNKNOWN` will be interpreted according to the defaults specific for a particular structure type.

carpawell (approved):

cthulhu-rider (approved):

PR #203, opened 2022-03-02
refs: Change support for different signature schemes

cthulhu-rider:
Remove `UNSPECIFIED` value from `SignatureScheme` enum. Make
`ECDSA_SHA512` to be default signature scheme (zero value).

Define `SignatureRFC6979` type for RFC-6979 signatures. Use it in
`Container` service.

Signed-off-by: Leonard Lyubich <leonard@nspcc.ru>

### DIFF
--- a/refs/types.proto
+++ b/refs/types.proto
@@ -105,18 +105,25 @@ message Signature {
   bytes key = 1 [json_name = "key"];
   // Signature
   bytes sign = 2 [json_name = "signature"];
-  // Scheme contains digital signature scheme identifier.
+  // Scheme contains digital signature scheme identifier
   SignatureScheme scheme = 3 [json_name = "scheme"];
 }
 
 // Signature scheme describes digital signing scheme used for (key, signature) pair.
 enum SignatureScheme {
-  // Unknown. The default interpretation depends on a particular structure type.
-  UNSPECIFIED = 0;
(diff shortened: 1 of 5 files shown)


refs/types.proto
```diff
@@ -105,18 +105,25 @@ message Signature {
   bytes key = 1 [json_name = "key"];
   // Signature
   bytes sign = 2 [json_name = "signature"];
-  // Scheme contains digital signature scheme identifier.
+  // Scheme contains digital signature scheme identifier
```
fyrchik, line 108: But why?

fyrchik (approved):

ZhangTao1596:
May I ask why we need `SignatureRFC6979` since we already have `ECDSA_RFC6979_SHA256` scheme in `Signature`?

alexvanin:
`Signature` message allows to specify any supported scheme. That implies that message can be signed using `ECDSA_RFC6979_SHA256` **or** `ECDSA_SHA512`. Both are okay.

It is not true for container creation operation. Container creation operation requires **only** `ECDSA_RFC6979_SHA256` (because signature is verified in smart contract, smart contract can't verify `ECDSA_SHA512`). 

To stress that in protocol, we have separate `SignatureRFC6979` message that, by definition, expects `ECDSA_RFC6979_SHA256` signature. This message is used in container service.

PR #204, opened 2022-03-05
Clarification for eXtended Headers in MetaHeader

realloc:
Minor clarification to show how XHeaders should be used.

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -24,8 +24,9 @@ service ObjectService {
   //   Will use the requsted version of Network Map for object placement
   //   calculation.
   // * __NEOFS__NETMAP_LOOKUP_DEPTH \
-  //   Will try older versions of network map to find an object until the depth
-  //   limit reached.
+  //   Will try older versions (starting from `__NEOFS__NETMAP_EPOCH` if specified or
+  //   the latest one otherwise) of Network Map to find an object until the depth
+  //   limit is reached.
   //
   // Please refer to detailed `XHeader` description.
   //
@@ -164,8 +165,8 @@ service ObjectService {
   //   Will use the requsted version of Network Map for object placement
   //   calculation.
   // * __NEOFS__NETMAP_LOOKUP_DEPTH \
-  //   Will try older versions of network map to find an object until the depth
-  //   limit reached.
+  //   Will try older versions of Network Map to find an object until the depth
+  //   limit is reached.
   //
   // Please refer to detailed `XHeader` description.
   //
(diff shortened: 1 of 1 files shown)


session/types.proto
```diff
@@ -117,8 +117,15 @@ message SessionToken {
...
 }
 
-// Extended headers for Request/Response. May contain any user-defined headers
-// to be interpreted on application level.
+// Extended headers for Request/Response may contain any user-defined headers to
+// be interpreted on application level. It may be used by Storage Nodes for
+// experimental protocol extensions with official adoption of such changes in
+// the future protocol versions.
+//
+// It may be useful to have extended headers in both Request and Response. For
```
alexvanin, line 125: This sentence is a bit confusing for me. It "**might be** useful" but, "Response **should have**". So is it strictly required by the spec or just a recommendation?

alexvanin (commented):
What about mentioning two well-known headers in corresponding RPC method docs?

session/types.proto
realloc: @EliChin we need help in fixing the language here. I wanted to say that it's logical to make response message have an indication of some extended header used during request processing. There may be cases when this makes no sense, like asking to log the request or similar "invisible" actions, so it's not a strict requirement.

object/service.proto
```diff
@@ -42,6 +52,13 @@ service ObjectService {
   // object payload. All messages, except first one, SHOULD be payload chunks.
   // Chunk messages SHOULD be sent in direct order of fragmentation.
   //
+  // Extended headers can change `Put` behaviour:
+  // * __NEOFS__NETMAP_EPOCH \
```
cthulhu-rider, line 56: I don't think this is useful for `Put`. I can't imagine the scenario when somebody wants to store an object in the "past" and storage nodes accepts it.

Same for `Delete`.

object/service.proto
```diff
@@ -19,6 +19,16 @@ service ObjectService {
   // be restored by concatenation of object message payload and all chunks
   // keeping receiving order.
   //
+  // Extended headers can change `Get` behaviour:
+  // * __NEOFS__NETMAP_EPOCH \
+  //   Will use the requsted version of Network Map for object placement
+  //   calculation.
+  // * __NEOFS__NETMAP_LOOKUP_DEPTH \
+  //   Will try older versions of network map to find an object until the depth
```
cthulhu-rider, line 27: I'd clarify:
```suggestion
  //   Will try older versions (starting from `__NEOFS__NETMAP_EPOCH` if specified or the latest one otherwise) of network map to find an object until the depth
```
in order to prevent infinite request loop if storage nodes always start processing from the latest epoch.

object/service.proto
realloc: This functionality is required to address cases with some network segments losing full connectivity. Even if it's not used now by our implementation, the protocol itself must support operation in an Epoch specified.

cthulhu-rider: I agree that we can parameterize this value in binary protocol, but I still can't imagine the UX in setting the particular values. Regardless of the implementation, how are storage nodes that are no longer present in the container motivated to serve these kinds of requests?

carpawell (approved):

cthulhu-rider (approved):

PR #216, opened 2022-04-21
English Check

EliChin:
Checked for some inaccuracies regarding English.

Signed-off-by: Elizaveta Chichindaeva <elizaveta@nspcc.ru>

### DIFF
--- a/acl/types.proto
+++ b/acl/types.proto
@@ -15,11 +15,11 @@ enum Role {
   // User target rule is applied if sender is the owner of the container
   USER = 1;
 
-  // System target rule is applied if sender is the storage node within the
-  // container or inner ring node
+  // System target rule is applied if sender is a storage node within the
+  // container or an inner ring node
   SYSTEM = 2;
 
-  // Others target rule is applied if sender is not user nor system target
+  // Others target rule is applied if sender is neither a user nor a system target
   OTHERS = 3;
 }
 
@@ -100,7 +100,7 @@ message EACLRecord {
   // Rule execution result. Either allows or denies access if filters match.
   Action action = 2 [json_name = "action"];
 
-  // Filter to check particular properties of the request or object.
+  // Filter to check particular properties of the request or the object.
   //
   // By default `key` field refers to the corresponding object's `Attribute`.
   // Some Object's header fields can also be accessed by adding `$Object:`
@@ -160,12 +160,12 @@ message EACLRecord {
   repeated Target targets = 4 [json_name="targets"];
 }
 
-// Extended ACL rules table. Defined a list of ACL rules additionally to Basic
-// ACL. Extended ACL rules can be attached to the container and can be updated
+// Extended ACL rules table. A list of ACL rules defined additionally to Basic
+// ACL. Extended ACL rules can be attached to a container and can be updated
 // or may be defined in `BearerToken` structure. Please see the corresponding
-// NeoFS Technical Specification's section for detailed description.
+// NeoFS Technical Specification section for detailed description.
 message EACLTable {
-  // eACL format version. Effectively the version of API library used to create
+  // eACL format version. Effectively, the version of API library used to create
   // eACL Table.
   neo.fs.v2.refs.Version version = 1 [json_name = "version"];
 
@@ -183,11 +183,11 @@ message EACLTable {
 // used in the similar use cases, like providing authorisation to externally
 // authenticated party.
 //
-// BearerToken can be issued only by container's owner and must be signed using
-// the key associated with container's `OwnerID`.
+// BearerToken can be issued only by the container's owner and must be signed using
+// the key associated with the container's `OwnerID`.
 message BearerToken {
-  // Bearer Token body structure contains Extended ACL table issued by container
-  // owner with additional information preventing token's abuse.
+  // Bearer Token body structure contains Extended ACL table issued by the container
+  // owner with additional information preventing token abuse.
   message Body {
     // Table of Extended ACL rules to use instead of the ones attached to the
     // container. If it contains `container_id` field, bearer token is only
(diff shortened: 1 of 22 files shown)


proto-docs/container.md
```diff
@@ -109,7 +109,7 @@ Statuses:
 | Delete | [DeleteRequest](#neo.fs.v2.container.DeleteRequest) | [DeleteResponse](#neo.fs.v2.container.DeleteResponse) |
 #### Method Get
 
-Returns container structure from `Container` smart contract storage.
+This method returns container structure from `Container` smart contract storage.
```
fyrchik, line 112: Simple `Returns` is quite common, especially if you render it as a table.

container/service.proto
```diff
@@ -47,16 +47,16 @@ service ContainerService {
...
   //
   // Statuses:
   // - **OK** (0, SECTION_SUCCESS): \
   //   container list has been successfully read;
   // - Common failures (SECTION_FAILURE_COMMON).
   rpc List(ListRequest) returns (ListResponse);
 
-  // Invokes 'SetEACL' method of 'Container` smart contract and returns response
-  // immediately. After one more block in sidechain, Extended ACL changes are
+  // This method invokes 'SetEACL' method of 'Container` smart contract and returns response
```
realloc, line 58: It's obvious, that we describe this method, hence was omitted initially.

object/service.proto
```diff
@@ -10,14 +10,14 @@ import "refs/types.proto";
 import "session/types.proto";
 
 // `ObjectService` provides API for manipulating objects. Object operations do
-// not interact with sidechain and are only served by nodes in p2p style.
+// not address sidechain and are only served by nodes in p2p style.
```
realloc, line 13: Not sure if it's the right verb here.

object/service.proto
```diff
@@ -116,9 +116,9 @@ service ObjectService {
   //   provided session token has expired.
   rpc Search(SearchRequest) returns (stream SearchResponse);
 
-  // Get byte range of data payload. Range is set as an (offset, length) tuple.
+  // Get byte range of data payload. Range is set as a (offset, length) tuple.
```
realloc, line 119: Here the '()' are defining the tuple, so `a` comes right before `o` and should be transformed into `an`, isn't it?

status/types.proto
```diff
@@ -5,7 +5,7 @@ package neo.fs.v2.status;
 option go_package = "github.com/nspcc-dev/neofs-api-go/v2/status/grpc;status";
 option csharp_namespace = "Neo.FileStorage.API.Status";
 
-// Declares the general format of the status returns of the NeoFS RPC protocol.
+// It declares the general format of the status returns of the NeoFS RPC protocol.
```
realloc, line 8: Let's use just the action verb in the beginning of a method description. It's a common practice, I think.

acl/types.proto
```diff
@@ -160,12 +160,12 @@ message EACLRecord {
   repeated Target targets = 4 [json_name="targets"];
 }
 
-// Extended ACL rules table. Defined a list of ACL rules additionally to Basic
-// ACL. Extended ACL rules can be attached to the container and can be updated
+// Extended ACL rules table. A list of ACL rules defined additionally to Basic
```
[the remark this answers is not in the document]
EliChin: Why? The meaning is that a list of extended ACls are defined in addition to basic ACL rules, isn't it? If we used `defines`, it would be like `a list defines` what? `additionally to basic rules`.

proto-docs/container.md
EliChin: `proto-docs` will be recovered to master

object/service.proto
EliChin: Well, I would not consider anything in brackets but check what stands after them, like `a ... tuple`.
https://www.quora.com/In-English-grammar-should-the-word-before-the-parenthesis-be-an-or-a-ie-do-you-match-it-to-the-first-word-in-the-bracket-or-first-word-after-the-closing-bracket

object/service.proto
EliChin: I just dislike `operations interact with`, I would change the verb after `operations`

EliChin:
`proto-doc` retrieved to the master state

reputation/types.proto
```diff
@@ -43,7 +43,7 @@ message PeerToPeerTrust {
 
 // Global trust level to NeoFS node.
 message GlobalTrust {
-    // Message format version. Effectively the version of API library used to create
+    // Message format version. The API library version effectively used to create
```
carpawell, line 46: i think there could happen a misunderstanding: the idea of the comment was: "That field means the version of that message. In fact (usually/in practice), the version of the API library (which was used to create that message) is used here."

please, check that your version brings the same meaning

object/types.proto
```diff
@@ -77,7 +77,7 @@ message ShortHeader {
 
 // Object Header
 message Header {
-  // Object format version. Effectively the version of API library used to
+  // Object format version. The API library version effectively used to
```
carpawell, line 80: same

object/types.proto
```diff
@@ -51,7 +51,7 @@ enum MatchType {
 
 // Short header fields
 message ShortHeader {
-  // Object format version. Effectively the version of API library used to
+  // Object format version. The API library version effectively used to
```
carpawell, line 54: same

container/types.proto
```diff
@@ -10,11 +10,11 @@ import "refs/types.proto";
 
 // Container is a structure that defines object placement behaviour. Objects can
 // be stored only within containers. They define placement rule, attributes and
-// access control information. ID of the container is a 32 byte long SHA256 hash
+// access control information. An ID of a container is a 32 byte long SHA256 hash
 // of stable-marshalled container message.
 message Container {
-  // Container format version. Effectively the version of API library used to
-  // create container.
+  // Container format version. The API library version effectively used to
```
carpawell, line 16: same

carpawell (approved):

reputation/types.proto
EliChin: Got it.
That's a good example why commas are very important sometimes.

PR #224, opened 2022-06-16
Fix/Clarify container removal

carpawell:


cthulhu-rider (approved):
LGTM

realloc:
Shouldn't we decide on container lock/hold procedure first?

carpawell:
> Shouldn't we decide on container lock/hold procedure first?

Yes, lets discuss it first and then correct that PR according to the decision.

realloc:
Note: In S3 API you can't remove a non-empty bucket, hence there is no such issue. In NeoFS we can't check container content on DELETE operation, so some additional container lock/hold mechanism is required.

Related: nspcc-dev/neofs-contract#247

roman-khimov:
The PR describes the way things work now and having this documentation is better than not having it. If the way things work is to change, we'll change the documentation accordingly.

PR #225, opened 2022-06-16
Feat/Add signature related status

### LINKED ISSUE #200: Add signature-related status codes
## Context

After #55 there are several signature schemes. As new schemes appear, network participants may not immediately support them, and therefore validate them. It would be useful to highlight this fact in the response status.

## Possible solution

Collective status code relating to signatures `SIGNATURE_VERIFICATION_FAIL`. Reasons in detail:

1. incorrect value;
2. unsupported scheme.

carpawell:
Closes #200.

### DIFF
--- a/status/types.proto
+++ b/status/types.proto
@@ -95,6 +95,9 @@ enum CommonFail {
     //  - [**0**] Magic number of the served NeoFS network (big-endian 64-bit
     //  unsigned integer).
     WRONG_MAGIC_NUMBER = 1;
+
+    // [**1026**] Signature verification failure.
+    SIGNATURE_VERIFICATION_FAIL = 2;
 }
 
 // Section of statuses for object-related operations.


realloc (approved):

status/types.proto
```diff
@@ -95,6 +95,13 @@ enum CommonFail {
     //  - [**0**] Magic number of the served NeoFS network (big-endian 64-bit
     //  unsigned integer).
     WRONG_MAGIC_NUMBER = 1;
+
+    // [**1026**] Signature verification failure.
+    // Details:
+    //  - [**0**] Reason (big-endian 16-bit unsigned integer):
+    //   - [**0**] Signature does not match.
```
fyrchik, line 102: Signature is invalid?
Also, what about failing to unmarshal public key?

carpawell: it was decided not to add details since it is not fully clear what format is ok. at least for now

/cc @cthulhu-rider @realloc

PR #228, opened 2022-07-27
Add netmap snapshot request

realloc:
Clients need an easy way of getting network map to get connection endpoints form Node's attributes.

Related: nspcc-dev/neofs-sdk-go#297


carpawell:
Should be synced with the control service: not to duplicate functionality.

carpawell (approved):

Issue #238, opened 2022-08-17
Add well-known `FilePath` object attribute description


realloc:
We have a well-known `FilePath` attribute used to set a full path to an object in an emulated directory structure.
It's used in HTTP and S3 protocol gateways, we need to specify it.

PR #239, opened 2022-08-26
Add FilePath object attribute

realloc:
FilePath attribute is already used by S3 and HTTP protocol gateways.
Also seen in [Gaspump](https://github.com/configwizard/gaspump-api/), NeoFS.Send and NeoFS.CDN.

Closes: #238

### DIFF
--- a/object/types.proto
+++ b/object/types.proto
@@ -135,6 +135,13 @@ message Header {
   //   Human-friendly name
   // * FileName \
   //   File name to be associated with the object on saving
+  // * FilePath \
+  //   Full path to be associated with the object on saving. Should start with a
+  //   '/' and use '/' as a delimiting symbol. Trailing '/' should be
+  //   interpreted as a virtual directory marker. If an object has conflicting
+  //   FilePath and FileName, FilePath should have higher priority, because it
+  //   is used to construct the directory tree. FilePath with trailing '/' and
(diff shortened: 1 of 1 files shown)


object/types.proto
```diff
@@ -135,6 +135,13 @@ message Header {
   //   Human-friendly name
   // * FileName \
   //   File name to be associated with the object on saving
+  // * FilePath \
+  //   Full path to be associated with the object on saving. Should start with a
+  //   '/' and use '/' as delimiting symbol. Trailing '/' should be interpreted
```
KirillovDenis, line 140: Why we want to have such limitation? In S3 we can create both `/some/path/to/object` and `some/path/to/object`

EliChin, line 140: as a delimiting symbol

EliChin (dismissed):

object/types.proto
realloc: It's a recommendation to simplify directory hierarchy construction and avoid problems with incorrect path mapping. For example, a typical http proxy rewrite rules block expects object path starting with '/'. Otherwise, it needs to try both possible variants. We want to avoid such situations.

KirillovDenis (approved):

masterSplinter01 (approved):

PR #242, opened 2022-09-14
Maintenance node state and status code

### LINKED ISSUE #237: Maintenance mode support
- Add `MAINTENANCE_MODE` status code. It should be returned on every public API call if a Node is in maintenance mode.
- Add `MAINTENANCE` [Node state](https://github.com/nspcc-dev/neofs-api/blob/4de2570e1a3dfc9633bbf02daf74753e7ba46b69/netmap/types.proto#L226).

fyrchik:
Close #237 .

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -232,6 +232,9 @@ message NodeInfo {
 
     // Network unavailable state
     OFFLINE = 2;
+
+    // Maintenance state
+    MAINTENANCE = 3;
   }
 
   // Carries state of the NeoFS node
--- a/status/types.proto
+++ b/status/types.proto
@@ -98,6 +98,9 @@ enum CommonFail {
 
     // [**1026**] Signature verification failure.
     SIGNATURE_VERIFICATION_FAIL = 2;
+
+    // [**1027**] Node is under maintenance.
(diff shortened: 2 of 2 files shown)


cthulhu-rider (approved):

carpawell:
@realloc, I wanted to pay attention to the fact that the current PR adds the status and the mode but does not mention them in any API method like we usually [do](https://github.com/nspcc-dev/neofs-api/blob/15eaf93ba1eb6947fb6e0927a1e6c430866cfa52/container/service.proto#L71-L78).

realloc:
@fyrchik: Agree with @carpawell. Could you please fix this?

carpawell:
Oh, I missed that the new status was added to the `SECTION_FAILURE_COMMON`. Well, it is already included in all our APIs then. The one question left: should a node respond with `NODE_UNDER_MAINTENANCE` on *every* API call? Even on `NetmapSnapshot`, and other non-storage-engine requests?

fyrchik:
I think maintenance can be much more broader that replacing a disk, so unless we want to fix the API later, it is easier to respond with this status to every RPC. Responding with success if we are under maintenance is not against the spec.

PR #245, opened 2022-09-20
Add a well-known attribute for additional addresses

fyrchik:
The task says "external" but wouldn't "internal" be better here? I mean for backwards compatibility we have a single address to which user connects and optional internal addresses to use for storage nodes.

Signed-off-by: Evgenii Stratonikov <evgeniy@nspcc.ru>

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -202,6 +202,10 @@ message NodeInfo {
   //   Node's continent name according to the [Seven-Continent model]
   //   (https://en.wikipedia.org/wiki/Continent#Number). Calculated
   //   automatically from `UN-LOCODE` attribute.
+  // * ExternalAddr
+  //   Node's preferred way for communications with external clients.
+  //   Clients SHOULD use these addresses if possible.
+  //   Must contain a comma-separated list of multi-addresses.
   //
   // For detailed description of each well-known attribute please see the
   // corresponding section in NeoFS Technical Specification.


cthulhu-rider:
Agree with internal: external endpoints are currently presented in `addresses` field.

carpawell:
Well, the idea as I got it: internal communications are something that any node always has, that is expected to be the "main" node's communication, that is usually kept unchanged and that is *always* present in every node's lifecycle.
Optionally, every node *may* specify a *preferable* channel for external clients. If it is present, it *does not affect* the "main" channel, and all the nodes should always use [addresses](https://github.com/nspcc-dev/neofs-api/blob/b4cdca989fc3171b1179b629425b75bcb27f68c1/netmap/types.proto#L136) in (Node-Node) connections (and check if there are any optional attrs only in emergency cases). I believe that netmap is more of an internal entity rather than an external one.

If we use attrs for internal channels, that "optionality" is not that clear: it changes the "main" Node-Node interaction and mixes communication with the old nodes that do not know that should pass their traffic through the new interface.

I think @realloc could help us here.

realloc:
It has to declare endpoints for external communication with clients. Internal communication relies on existing mechanism of address declaration in netmap.

realloc (changes requested):
It has to describe addresses for external communication. Applications and gates should use those addresses.

realloc:
According to current behviour, nodes get peer address from the network map entries. Client software either does the same or uses endpoint set in config.

If a node admin wants to use separate interfaces for serving clients and for p2p communication with other storage nodes, he sets one more listening address in node config and uses it as connection endpoints in clients (protocol gateways, applications, etc).

Adding well known attributes for additional endpoint could simplify this process. Client libraries used in, for example, protocol gateways could correctly discover and use proper endpoints without mixing client traffic with internal replication traffic.

carpawell (approved):

cthulhu-rider (approved):

realloc (approved):

PR #246, opened 2022-09-23
netmap: Define keys of `NetworkConfig.Parameter`

### LINKED ISSUE #214: Define keys to the network config values
In current protocol network configuration is represented by list of raw KV pairs. When new net config is added, there is no way to select its value by key. I propose to add constant keys to the existing config values.

cthulhu-rider:
* closes #214

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -252,7 +252,47 @@ message Netmap {
 
 // NeoFS network configuration
 message NetworkConfig {
-  // Single configuration parameter
+  // Single configuration parameter. Key MUST be network-unique.
+  //
+  // System parameters:
+  // - **AuditFee** \
+  //   Fee paid by the storage group owner to the Inner Ring member.
+  //   Value: little-endian integer. Default: 0.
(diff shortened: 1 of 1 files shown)


realloc (approved):

netmap/types.proto
```diff
@@ -252,7 +252,46 @@ message Netmap {
...
+  //   Value: little-endian integer. Default: 0.
+  // - **ContainerAliasFee** \
+  //   Fee paid for named container's creation by the container owner.
+  //   Value: little-endian integer. Default: 0.
+  // - **ContainerFee** \
+  //   Fee paid for container creation by the container owner.
+  //   Value: little-endian integer. Default: 0.
+  // - **EigenTrustAlpha** \
+  //   Alpha parameter of EigenTrust algorithm used in the Reputation system.
+  //   Value: decimal floating-point number. Default: 0.
```
fyrchik, line 273: in UTF-8 string representation?

fyrchik (dismissed):

fyrchik (approved):

PR #248, opened 2022-10-26
Actualize documentation for ContainerService.List

acid-ant:
Relates to https://github.com/nspcc-dev/neofs-node/issues/1689

Signed-off-by: Anton Nikiforov <an.nikiforov@yadro.com>


realloc (changes requested):
We NEVER return full list of all containers in the network. Only filtered by specific owner.

PR #267, opened 2023-08-08
feat/Add version node attribute

carpawell:
Relates https://github.com/nspcc-dev/neofs-node/issues/2455.

@roman-khimov, @AliceInHunterland, @cthulhu-rider let's discuss the details:
1. Should it have `__NEOFS__` prefix?
2. Should we add some preferable scheme? e.g. "you should add implementation language, build time, and version", etc


cthulhu-rider:
i'd start with unformatted attribute (otherwise dedicated field would be much more clear and easy-to-support)

roman-khimov:
1. No
2. No languages, no build times, just version. If you mean that we can have multiple implementations, that's true, but they would all have some names. So just a string (OK, ASCII printable) is OK.

carpawell:
Ready for review then.

cthulhu-rider (approved):

Issue #269, opened 2023-08-09
Add `400 Bad Request`-like status code


cthulhu-rider:
example:
```
status: code = 1024 message = malformed request: invalid session token owner
```

i propose to extend https://github.com/nspcc-dev/neofs-api/blob/953b4c5a9df1409da2a795b34bf9af66699689c6/status/types.proto#L87 enum with  `BAD_REQUEST` value. The status will be returned to any request with incorrect format. It will allow to distinguish such failures from real internal server errors

carpawell:
@cthulhu-rider, do you distinguish:
1. Completely strange request (missing required fields, or container/object ID is longer than 32 bytes, etc)
2. Logically incorrect: your example (it is a correct request but its `owner` must not be treated that way)?

Or they both should be `400`ed the same way?

To me `1.` is about "you made a mistake in your client implementation" while `2.` is about "your admin made a mistake when providing a path to a wallet". I like details but think that for the most cases it does not matter.

cthulhu-rider:
@carpawell IIUC ur example is about diff b/w `400` and `401/403`. These are completely diff cases for sure

original example may mean both cases. I propose to clarify for better system responsiveness

i mean `400` bout incorrect fields' format or message sequences

cthulhu-rider:
server-side streams, which are object GET/RANGE/SEARCH, may break of due when communication with subsequent node fails. There is no code for this now, the server is forced to return `INTERNAL`

we can add `UNAVAILABLE` similar to the [gRPC](https://grpc.io/docs/guides/status-codes/) one. Note that verbally `ABORTED` looks closer, but iiuc it's about other cases. Note that we aint obliged to completely follow the gRPC naming

client profit:
- he can perform retries that should not be done on `INTERNAL`
- for GET/RANGE he can even continue: call RANGE for the remaining payload

cthulhu-rider:
boost this issue :arrow_up: 
and all the related ones

Issue #201, opened 2022-02-24
Stream `EOF` support


cthulhu-rider:
Currently we don't have any mention of a fair end to the stream. Go `gRPC` implementation returns `EOF` on client/server side if server/client closed the stream via corresponding library functionality.

Do we want to unify this behavior on protocol definition level? Possible solutions:

1. Require to finish stream with empty request|response body: this will support the completion of the stream at a purely protocol level, and not be tied to the implementation;
2. Require to finish stream with `EOF`-like status code: alternative way of 2.

cthulhu-rider:
i thought that the synonymous problem could be lack of `unknown RPC`: Protocol Buffers v3 doesnt contain functionality for this. Some dedicated status message would help

cthulhu-rider:
consider specific status code similar to [`206 Partial Content`](https://datatracker.ietf.org/doc/html/rfc2616#section-10.2.7). Then all non-fin messages could carry this code while the fin one `0 OK` meaning `EOF`

but be careful, backward compatibility never sleeps

roman-khimov:
Meta header could have an additional field for this. GRPC streaming isn't designed for our use case, so an additional field to separate premature stream closure can be helpful.

PR #280, opened 2023-12-11
object: Separate API for node-to-node replication

cthulhu-rider:
here is a design of the separate object replication API

in addition to general simplification, the main motivation is to reduce amount of transmitted service data (request headers, signatures, stream markup, etc.) compared to general purpose `rpc Put`

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -8,6 +8,7 @@ option csharp_namespace = "Neo.FileStorage.API.Object";
 import "object/types.proto";
 import "refs/types.proto";
 import "session/types.proto";
+import "status/types.proto";
 
 // `ObjectService` provides API for manipulating objects. Object operations do
 // not affect the sidechain and are only served by nodes in p2p style.
@@ -218,6 +219,23 @@ service ObjectService {
   // - **TOKEN_EXPIRED** (4097, SECTION_SESSION): \
   //   provided session token has expired.
   rpc GetRangeHash(GetRangeHashRequest) returns (GetRangeHashResponse);
+
+  // Save replica of the object on the NeoFS storage node. Both client and
+  // server must authenticate NeoFS storage nodes matching storage policy of
+  // the container referenced by the replicated object. Thus, this operation is
+  // purely system: regular users should not pay attention to it but use Put.
+  //
+  // Statuses:
+  // - **OK** (0, SECTION_SUCCESS): \
+  //   the object has been successfully replicated;
+  // - **INTERNAL_SERVER_ERROR** (1024, SECTION_FAILURE_COMMON): \
+  //   internal server error described in the text message;
+  // - **ACCESS_DENIED** (2048, SECTION_OBJECT): \
+  //   the client does not authenticate any NeoFS storage node matching storage
+  //   policy of the container referenced by the replicated object
+  // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER): \
+  //   the container to which the replicated object is associated was not found.
+  rpc Replicate(ReplicateRequest) returns (ReplicateResponse);
 }
 
 // GET object request
@@ -687,4 +705,20 @@ message GetRangeHashResponse {
   // authenticate the nodes of the message route and check the correctness of
   // transmission.
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
+}
+
+// Replicate RPC request
+message ReplicateRequest {
+  // Object to be replicated.
+  neo.fs.v2.object.Object object = 1;
+
+  // Signature of all other request fields serialized in Protocol Buffers v3
+  // format in ascending order of fields.
+  neo.fs.v2.refs.Signature signature = 2;
+}
+
+// Replicate RPC response
+message ReplicateResponse {
+  // Operation execution status with one of the enumerated codes.
+  neo.fs.v2.status.Status status = 1;
 }
--- a/proto-docs/object.md
+++ b/proto-docs/object.md
@@ -36,6 +36,8 @@
     - [PutResponse](#neo.fs.v2.object.PutResponse)
     - [PutResponse.Body](#neo.fs.v2.object.PutResponse.Body)
     - [Range](#neo.fs.v2.object.Range)
+    - [ReplicateRequest](#neo.fs.v2.object.ReplicateRequest)
+    - [ReplicateResponse](#neo.fs.v2.object.ReplicateResponse)
     - [SearchRequest](#neo.fs.v2.object.SearchRequest)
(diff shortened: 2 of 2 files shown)


object/service.proto
```diff
@@ -218,6 +219,9 @@ service ObjectService {
   // - **TOKEN_EXPIRED** (4097, SECTION_SESSION): \
   //   provided session token has expired.
   rpc GetRangeHash(GetRangeHashRequest) returns (GetRangeHashResponse);
+
+  // TODO: docs
+  rpc Replicate(ReplicationRequest) returns (ReplicationResponse);
```
roman-khimov, line 224: Will it work for objects larger than a single grpc message?

object/service.proto
```diff
@@ -688,3 +692,15 @@ message GetRangeHashResponse {
   // transmission.
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
 }
+
+// TODO: docs
+message ReplicationRequest {
+  // TODO: reserve 1st byte for forward compatibility
+  bytes object = 1;
+
+  neo.fs.v2.refs.Signature object_signature = 2;
```
roman-khimov, line 701: That's the node signature, right? Because there is already one in `Object`.

object/service.proto
cthulhu-rider: yes i've tested this within https://github.com/nspcc-dev/neofs-node/pull/2674/commits/62d020c8e62b805071bc7f9742b28d0c4129c60a

object/service.proto
cthulhu-rider: yeah right, we can improve naming, e.g. `object_field_signature`

object/service.proto
```diff
@@ -688,3 +692,15 @@ message GetRangeHashResponse {
   // transmission.
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
 }
+
+// TODO: docs
+message ReplicationRequest {
+  // TODO: reserve 1st byte for forward compatibility
+  bytes object = 1;
```
roman-khimov, line 699: Can we have it be structured then? Like a proper `Object`? That's the one we send anyway, so if you have it serialized, that's your message, or not?

cthulhu-rider: yeah all messages are binary

cthulhu-rider:
@roman-khimov ive finalized format and docs, take a look pls

object/service.proto
```diff
@@ -688,3 +695,37 @@ message GetRangeHashResponse {
...
+
+  // Signature of all other request fields serialized in Protocol Buffers v3
+  // format in ascending order of fields.
+  neo.fs.v2.refs.Signature signature = 2;
+}
+
+// Replicate RPC response
+message ReplicateResponse {
+  // Operation status codes.
+  enum Code {
```
roman-khimov, line 712: No `ResponseMetaHeader`? No `ResponseVerificationHeader`? No common codes? I'd expect it at least to follow common status data and reuse existing codes (`ResponseMetaHeader`). Separate codes can also be very inconvenient.

cthulhu-rider: predictable questions. For now, I only doubt the need to add a signature similar to the one in the request. The rest is overhead for this simple operation

roman-khimov: Unified error codes are very valuable to me.

cthulhu-rider: will reuse, not yet supported cases can fallback to `INTERNAL` (like #269)

cthulhu-rider: check pls

roman-khimov: We can extend the set if needed.

object/service.proto
```diff
@@ -688,3 +706,19 @@ message GetRangeHashResponse {
...
+
+  // Signature of all other request fields serialized in Protocol Buffers v3
+  // format in ascending order of fields.
+  neo.fs.v2.refs.Signature signature = 2;
+}
+
+// Replicate RPC response
+message ReplicateResponse {
+  // Operation execution status with one of the enumerated codes.
+  neo.fs.v2.status.Status status = 1;
```
roman-khimov, line 723: I like the simplicity of it, but this breaks https://github.com/nspcc-dev/neofs-sdk-go/blob/master/client/response.go and associated code around it. Does it affect integration?

object/service.proto
cthulhu-rider: there is no urgent need now

object/service.proto
cthulhu-rider: nope, integration is ok cuz this is purely new functionality. If protocol doesnt require signature (if we keep the current status-only response), then applications must not too

the only tricky thing i see is https://github.com/nspcc-dev/neofs-api/issues/201#issuecomment-1891383454, but there are app-side workarounds

the main thing for us is to think carefully about whether we need a signature from the protocol pov. And ofc, when in doubt, it is better to add it and then make it optional, because further additions must be optional for the backward compat sake. But, at the same time, the benefits are questionable while the resource costs are real

im leaning more towards the option without a signature as more efficient, but I'm not completely sure yet. Share ur thoughts pls @roman-khimov @carpawell

roman-khimov: Signatures make some sense for insecure connections or for rerouted messages, but for TLS-enabled 1:1 calls they're irrelevant. And then there is also some benefit in having insecure unsigned option for fast communication over known good channels. So if this inconsistency is not an obstacle, we better go without signatures.

Now for the other data like epochs, is it important for us to get it with _every_ reply? I doubt that, but maybe I'm missing something. This data costs nothing to transmit/receive.

cthulhu-rider: same thoughts about transport security. Can we spread this to requests also? In proposed changes, replication requests are signed while responses are not

> to get it with _every_ reply

absolutely not. For me, this is redundancy useful for a narrow set of tasks (mostly system ones), which should not be unconditionally transferred. I see the behavior: if client set flag in the request, epoch/smth will be returned, not by default. Although the amount of data like epoch is insignificant, it is still pointless to transfer it when not needed by the client

object/service.proto
```diff
@@ -688,3 +706,19 @@ message GetRangeHashResponse {
...
+  neo.fs.v2.object.Object object = 1;
+
+  // Signature of all other request fields serialized in Protocol Buffers v3
+  // format in ascending order of fields.
+  neo.fs.v2.refs.Signature signature = 2;
+}
+
+// Replicate RPC response
+message ReplicateResponse {
+  // Operation execution status with one of the enumerated codes.
```
carpawell, line 722: i would say it is an incorrect comment now. i see `Status`, no enumeration

object/service.proto
carpawell: i think a signature absence is acceptable for a system-only RPC response

roman-khimov (approved):

Issue #264, opened 2023-05-30
Make it easy to seek through large objects


roman-khimov:
There are several ways to do this:
1. Store the size of each chunk in the link object along with the hash. It has the most flexibility, but it also means additional per-part overhead to store this.
2. Keep link object as is, but store the offset from the beginning in each regular object and at least allow to bisect into some particular offset instead of looping through the whole list (10 HEADs is much better than 1000 HEADs).
3. Demand individual parts to be of the same size, specify this size in the split data header. It's pretty easy to do, we probably HEAD `header.split.previous` anyway, so checking that the size of the object is the same as the previous one costs nothing.

carpawell:
I vote for the first one as the most explicit object-handling information to me: helper object has helper info. An offset inside every object seems redundant info for the object itself, it does not take part in the assembly and not required for the storing payload. Demanding the same size would be the best and the easiest to me but we have not done it yet and it looks like a feature, mb we do not want to be so strict.

roman-khimov:
Storing offset in a small object costs nothing, storing sizes in the link object does have some cost to it (~10% per-object additional overhead). We can do both though.

Equal parts are not compatible with non-reslicing S3 multipart handler (https://github.com/nspcc-dev/neofs-s3-gw/issues/843).

PR #283, opened 2024-01-12
fix/Child objects ACL

carpawell:


object/types.proto
```diff
@@ -29,6 +29,9 @@ enum ObjectType {
 
   // Object lock
   LOCK = 3;
+
+  // Initialization part of the split regular object
+  INIT = 4;
```
roman-khimov, line 34: So, a regular object with regular meta, but just a different type, right?

carpawell: yes. we can discuss if there are any other worthwhile options. also, we can think if the link object differs from the suggested init object and if we should make it special too

roman-khimov:
Let's make a complete spec for the new format first, then implement it step by step fixing along the way if needed.

carpawell:
A few intermediate questions:

1. Should the link object be a separate object type? There are no requirements (at least for now) for it. Making a separate type for it certainly makes it harder to adopt in the node and may increase the number of unexpected bugs. Although, I just _feel_ like it should be a separate object type.
2. Is it OK to have a "broken" parent header in the initial object? It just cannot have any correct payload-dependent fields. 
3. Do we need some structure for the initial object too? Should it have a payload or parent object is ok to be in the header like it was before?

My answers:
1. Yes (but be ready it takes much more time to implement)
2. No
3. Not sure. Want to say "yes" since it is clearer to me but it would be too many different messages and types IMO (especially with `2.`).

roman-khimov:
1. A separate type likely will make it easier to adopt, you'll be able to separate the old one from the new easily.
2. It can have a lot of data in many cases, what's broken about it? Hashes?
3. It has a special type, then it depends on what else we need. Payload should OK.

carpawell:
> 1. A separate type likely will make it easier to adopt

If we make it from scratch -- yes. But currently a lot of places think that link object is a regular object. I can not even imagine how many. Moreover, backward compatibility is a question here: there are places that are not ready to get anything except REGULAR, TOMB, LOCK. But I do support you.

> 2. It can have a lot of data in many cases, what's broken about it? Hashes?

No ID, no payload length (in a general case), no hashes. That is not a header I would say. Previously, "parent" was a ready term in NeoFS. If you saw "parent" somewhere, you know what it was and were able to get anything you want about it. But not finished header solves our problem for now.

> 3. It has a special type, then it depends on what else we need. Payload should OK.

So you think `INIT` object should have a payload? Or not? Currently it does not have it and parent is placed in the header's `split.parent` field. It if is a separate message in its payload, we may think about storing parent header not like a [header](https://github.com/nspcc-dev/neofs-api/blob/8eb21aac3f6a811016536751a1b6c4d8f02c0428/object/types.proto#L79) but like a new struct without "payload" fields.

object/types.proto
```diff
@@ -181,6 +185,8 @@ message Header {
     // the same `split_id` value.
     bytes split_id = 6 [json_name = "splitID"];
 
+    // Identifier of the initial part of the object split chain
```
cthulhu-rider, line 188: required?

object/types.proto
```diff
@@ -222,4 +228,9 @@ message SplitInfo {
   // split header with the original object header and a sorted list of
   // object parts.
   neo.fs.v2.refs.ObjectID link = 3;
+
+  // The identifier of the initial object for split hierarchy parts. It contains
+  // zero payload and the original object (also known as a "parent") header in the
```
cthulhu-rider, line 233: known by who? all other fields and `SplitInfo` itself use `original object` term, and it's enough imo

link/types.proto
```diff
@@ -0,0 +1,20 @@
...
+// Link is a helper object that contains the full list of the split chain objects'
+// IDs. It is created only after the whole split chain is known and signed. This
+// object is the only object that has access to every "child object" ID. It is NOT
+// required for the original object assembling. It MUST have ALL the "child objects"
+// IDs. IDs MUST be ordered according to the original payload split, meaning the
+// first payload part holder MUST be placed at the first place in the corresponding
+// link object.
+message Link {
+  // Full list of the "child" object IDs.
+  repeated neo.fs.v2.refs.ObjectID children = 1 [json_name = "children"];
```
cthulhu-rider, line 19: how about cross-ref with https://github.com/nspcc-dev/neofs-api/blob/d95228c40283cf6e188073a87a802af7e5dc0a7d/object/types.proto#L179?

link/types.proto
```diff
@@ -0,0 +1,20 @@
...
+import "refs/types.proto";
+
+// Link is a helper object that contains the full list of the split chain objects'
+// IDs. It is created only after the whole split chain is known and signed. This
+// object is the only object that has access to every "child object" ID. It is NOT
+// required for the original object assembling. It MUST have ALL the "child objects"
+// IDs. IDs MUST be ordered according to the original payload split, meaning the
+// first payload part holder MUST be placed at the first place in the corresponding
+// link object.
+message Link {
```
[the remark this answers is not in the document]
carpawell: depends on the final solution about a separate object type. if it is a separate object type, i do not see any difference with `tombstone`, `storagegroup` and `lock` pkgs. but if you ask me, should it be like that, my answer is no, but we have what we have, do not want object type to be different in any way

link/types.proto
```diff
@@ -0,0 +1,20 @@
...
+
+package neo.fs.v2.link;
+
+option go_package = "github.com/nspcc-dev/neofs-api-go/v2/link/grpc;link";
+option csharp_namespace = "Neo.FileStorage.API.Link";
+
+import "refs/types.proto";
+
+// Link is a helper object that contains the full list of the split chain objects'
+// IDs. It is created only after the whole split chain is known and signed. This
```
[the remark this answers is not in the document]
carpawell: i agree that some statements are too wordy but i had read object API before i started this work and i was sad about a reader who only has this repo and tries to implement his own object split. so IMO too many words are better than too few

link/types.proto
```diff
@@ -0,0 +1,20 @@
...
+// object is the only object that has access to every "child object" ID. It is NOT
+// required for the original object assembling. It MUST have ALL the "child objects"
```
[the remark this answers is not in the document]
carpawell: dont have a strong opinion on this. maybe. but it seems that `object` pkg prefers an explanation for the message declaration over for a field one

object/types.proto
```diff
@@ -29,6 +30,9 @@ enum ObjectType {
 
   // Object lock
   LOCK = 3;
+
+  // Initialization part of the regular but split object
```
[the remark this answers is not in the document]
carpawell: agree, dropped

object/types.proto
carpawell: the field? it gonna be the only thing that saves us from ACL rules based on parent meta info

or you mean mentioning that this field is required. are the other onec required?

object/types.proto
carpawell: > known by who?

by an api reader only a few lines [above](https://github.com/nspcc-dev/neofs-api/blob/8eb21aac3f6a811016536751a1b6c4d8f02c0428/object/types.proto#L165). again, API docs are not perfect so more words are better imo

link/types.proto
carpawell: you mean comment ref? or what?

cthulhu-rider:
> separate type link

the concept is very simple: if payload format is strict, then separate type is required. Ofc some servers could deny service of unsupported types, so `LINK` objects must be essentially optional within the current protocol version. It’s nice that link objects in the current scheme are optional, this will allow both of its variations to be present in the same network

> Do we need some structure for the initial object too?

it cannot do without logical features expressed in structural differences. For example, there should be no incompatible collisions with the current split [schema](https://github.com/nspcc-dev/neofs-spec/blob/master/01-arch/04-object_split.md). About the structure itself, i think it should be as efficient as possible to solve exact problems regardless of the server implementation

link/types.proto
```diff
@@ -0,0 +1,30 @@
...
+
+import "refs/types.proto";
+
+// Link is a payload of helper objects that contain the full list of the split
+// chain objects' IDs. It is created only after the whole split chain is known
+// and signed. This object is the only object that refers to every "child object"
+// ID. It is NOT required for the original object assembling. It MUST have ALL
+// the "child objects" IDs. Child objects MUST be ordered according to the
+// original payload split, meaning the first payload part holder MUST be placed
+// at the first place in the corresponding link object. Sizes MUST NOT be omitted
```
carpawell, line 16: if we wanna use it, we should be able to rely on it so there is MUST word. but then, we need to validate it and that is N head requests per link object PUT (N is num of objects in the chain)

object/types.proto
```diff
@@ -29,6 +30,9 @@ enum ObjectType {
 
   // Object lock
   LOCK = 3;
+
+  // Initialization part of the but split object
+  INIT = 4;
```
carpawell, line 35: no payload for init objects for now. have not found requirements in it, info form the `parent_header` field is enough

link/types.proto
cthulhu-rider: > N head requests per link object

the problem is more specific - when assembling a split object, the list of descriptors of child objects can diverge with the stored children (not only the sizes btw, IDs too). The proposed validation is correct, but since we cannot guarantee it at the protocol level (this is a server implementation detail), the assembler - one who operates with split chain - must be prepared for the divergence in any case

protocol-level solution is to additionally store such metadata in the child object headers themselves (mentioned [here](https://github.com/nspcc-dev/neofs-api/issues/264#issuecomment-1908432474)). Then the link object will remain a pure stand-alone helper

link/types.proto
```diff
@@ -0,0 +1,30 @@
...
+  // Object ID with its object's payload size.
+  message MeasuredObject {
+    // Object ID.
+    neo.fs.v2.refs.ObjectID id = 1 [json_name = "id"];
+
+    // Object size in bytes.
+    uint32 size = 2 [json_name = "size"];
+  }
+
+  // Full list of the "child" object IDs.
```
cthulhu-rider, line 28: not IDs anymore
```suggestion
  // Full list of the "child" object descriptors.
```

object/types.proto
```diff
@@ -174,13 +178,17 @@ message Header {
     Header parent_header = 4 [json_name = "parentHeader"];
 
     // List of identifiers of the objects generated by splitting current one.
+    // Emtpy children list in a link object means that children's IDs
```
cthulhu-rider, line 181: here `link object` should refer to `SplitInfo.link` cuz that is where this entity is defined. This part is learned much earlier

object/types.proto
cthulhu-rider: > or you mean mentioning that this field is required

so, answer is "yes"

imo it’s worth explicitly pointing out that the field is present in all parts except the `INIT` itself (this is understandable, but not obvious)

object/types.proto
```diff
@@ -29,6 +30,9 @@ enum ObjectType {
 
   // Object lock
   LOCK = 3;
+
+  // Initialization part of the but split object
```
cthulhu-rider, line 34:
```suggestion
  // Initialization part of the split object
```

link/types.proto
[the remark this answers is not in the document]
cthulhu-rider: imo the previously used approach only complicates the structuring and, as a consequence, the knowledge of the protocol concepts. All these entities are inextricably linked with objects - they are about objects, they are in the payload of objects. Сurrent packetization is extremely redundant to me

> we have what we have

obviously, but we are not burdened with creating a new package (). Keep the existing ones, and this message can be left in `object` to comply with the system architecture hierarchy

being a newbie, i'll see `acl`, `container`, `object` and other large sections of the system. And then `link` - specific detail elevated to the top level of the system with a non-descript name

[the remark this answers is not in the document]
carpawell: > to comply with the system architecture hierarchy

so do i. type of the objects are placed where they always have been

the issue was created https://github.com/nspcc-dev/neofs-api/issues/284

object/types.proto
carpawell: > imo it’s worth explicitly pointing out that the field is present in all parts

agree, added some words

link/types.proto
carpawell: > but since we cannot guarantee it at the protocol level (this is a server implementation detail)

what do you mean? nodes may not accept such link objects

> must be prepared for the divergence in any case

there is no solution for #264 then. if you cannot trust the link object, you cannot be sure you are answering with a correct range

> protocol-level solution is to additionally store such metadata in the child object headers themselves

should also be validated somehow. a node can get Nth object with some offset in it. what does it do? just believes it is true? searches for the previous ones and calculates their sizes?

@roman-khimov

object/types.proto
carpawell: clarify your thoughts. `children` field only relates the Link object. a reader should have understood it somehow before. what do you mean by "should refer to SplitInfo.link"?

cthulhu-rider: > children field only relates the Link object.

i dont see this in `children` field comment. In current description this field has no connection with linking objects

> a reader should have understood it somehow before

and if he dont? Usually i see ref that allows me to become prepared

> what do you mean by "should refer to SplitInfo.link"

any ref that will allow reader to know _why_ this field is actually needed. Ref to `SplitInfo.link` will make it clear that the field is present in and only in linking objects

carpawell: > i dont see this in children field comment

it always was that way and the comment was not clear always

>> a reader should have understood it somehow before
>
> and if he dont?

then it is sad. like was before this PR

>  Ref to SplitInfo.link

what is ref to smth? like in go pkg? with `[]`? i did not find any examples: 
```
▶ grep -r -E "// .+\[" .
./acl/types.proto:// like [JWT](https://jwt.io), it has a limited lifetime and scope, hence can be
./netmap/types.proto:  //   [UN/LOCODE](https://www.unece.org/cefact/codesfortrade/codes_index.html)
./netmap/types.proto:  //   [ISO 3166-1_alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
./netmap/types.proto:  //   [ISO-3166](https://www.iso.org/obp/ui/#search). Calculated automatically
./netmap/types.proto:  //   Presented in [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2)
./netmap/types.proto:  //   [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2). Calculated
./netmap/types.proto:  //   Node's continent name according to the [Seven-Continent model]
./accounting/types.proto:// Please see [The General Decimal Arithmetic
./refs/types.proto:// String presentation is a [Base58
./status/types.proto:    //  - [**0**] Magic number of the served NeoFS network (big-endian 64-bit
./status/types.proto:    // - [**0**] Human-readable description (UTF-8 encoded string).
./reputation/service.proto:        // floating point number in [0;1] range.
./reputation/types.proto:    // Trust level in [0:1] range
```

link/types.proto
cthulhu-rider: > what do you mean? nodes may not accept such link objects

nodes may do whatever they want. They may accept linking objects w/o any validation

> there is no solution for https://github.com/nspcc-dev/neofs-api/issues/264 then

i dont deny current solution, im just highligting that linking object _can_ be unavailable or diverge with the split chain and any network node must be ready for this. In these cases, extra metadata in child objects _may_ be a painkiller in some cases

> what does it do?

its own implementation detail, each node can only control what to do itself and what it receives from other nodes

---

or maybe i misunderstood ur original comment. I dont see any problem with current structure and requirements. If u didnt mean that `N head requests per link object PUT` is a _problem_, then everything is ok to me

object/types.proto
cthulhu-rider: > it always was that way and the comment was not clear always

that's why i suggested clarifications - to make it clearer

> what is ref to smth?

reference

cthulhu-rider (dismissed):
format is ok, left some discussions about docs

object/types.proto
carpawell: > reference

i meant, how do you see it? how should it look like?

link/types.proto
carpawell: > nodes may do whatever they want
>> what does it do?
>
> its own implementation detail, each node can only control what to do itself and what it receives from other nodes

i meant we need to find a balance to implement our "perfect" node that may exist in the world of "worst" nodes. if we may not rely on link object, i think the whole work is useless. but if we want to rely on it, we should validate it somehow

> extra metadata in child objects may be a painkiller in some cases

here i want to say that ensuring child object data is ok is the same to ensuring similar info about the link object

> If u didnt mean that N head requests per link object PUT is a problem, then everything is ok to me

yes. this and the consequences

object/types.proto
cthulhu-rider: plain text

btw
```suggestion
    // Empty children list in a link object means that children's IDs
```

object/types.proto
roman-khimov: Allowing payload can make it more efficient. At the same time, likely this can be extended in future.

link/types.proto
roman-khimov: Validation should be implemented. Then I'd be optimistic wrt the link object contents. If it's not correct --- either you get an error (trying to access inexisting object) or reply with a wrong result, but it's not node's fault, garbage in -- garbage out.

roman-khimov (commented):
Can we have a behavior spec in https://github.com/nspcc-dev/neofs-spec/? I guess most of mechanics stays the same, except for INIT.

I'm also not sure how split_id/init interact, do we need split_id if init is present?

object/types.proto
carpawell: fixed typo

> plain text

well, i tried. i dont know what a ref is in our API. we haven't done it yet

carpawell:
> Can we have a behavior spec in https://github.com/nspcc-dev/neofs-spec/?

You mean before this PR merge? Let's have at least approve+draft to ensure we agree about this PR. I can't continue before we are done here, it has already changed 3 times.

> I'm also not sure how split_id/init interact, do we need split_id if init is present?

I'm not sure init object changes something here: if the link and the init objects are lost _and_ there is no SplitID, there is no ability to assembly the original object (except you find _somehow_ the latest part and compare its parent header). Also, you cannot update any indexes related to the original object if you meet some object part before the init or the link object (possible situation, why not?). cc @cthulhu-rider

object/types.proto
carpawell: > Allowing payload can make it more efficient.

Only if it has a separate message without useless (unknown for the init object) fields. But on the other hand, it breaks our rule: metabase is for header info only (but the link object has already done it too in this PR).

link/types.proto
carpawell: > Validation should be implemented.

Yes, sure. Just not sure about some link object replication: imagine you have a 1к parts object and receive a link to it for the first time: you need to do 1k head objects before you finish with this object. And you may be an attacker that sends 1k of such 1k-parts objects. 64mb data for a HEAD chaos in the NeoFS network.

roman-khimov (dismissed):

object/types.proto
roman-khimov: > metabase is for header info only

I'm not sure what's the conflict. Header goes into meta, payload is not needed there. Link object is a different story.

carpawell:
- no `INIT` object, the first part is used as a parent object's header holder
- `LINK` object is a separate type
- `SplitID` is deprecated, the first part is used instead (so it is placed in every part now)

roman-khimov (approved):

cthulhu-rider (approved):

cthulhu-rider:
@carpawell closes #264?

carpawell:
> closes https://github.com/nspcc-dev/neofs-api/issues/264?

Yes, added it to one of the commits.

PR #285, opened 2024-01-29
object: Support numerical search queries

### LINKED ISSUE #265: Numeric search matches
Similar to #255, but for Search matches, they have a `MatchType` of their own. Yay!

cthulhu-rider:
Previously, protocol did not support numerical search queries for objects. This may be needed for system attributes such as payload size or creation epoch, and for user ones if required by the client application.

New values of `MatchType` enumeration are added: `>`, `>=`, `<`, `<=`. These attributes allow user to search for objects with any decimal attributes. While only base-10 numbers are allowed, additional bases may be supported in the future without new enumerations

Note that the previously added `STRING_EQUAL` / `STRING_NOT_EQUAL` matchers are suitable for numerical `==` / `!=`.

Closes #265.

### DIFF
--- a/object/types.proto
+++ b/object/types.proto
@@ -47,6 +47,18 @@ enum MatchType {
 
   // String prefix match
   COMMON_PREFIX = 4;
+
+  // Numerical 'greater than'
+  NUM_GT = 5;
+
+  // Numerical 'greater or equal than'
+  NUM_GE = 6;
+
+  // Numerical 'less than'
+  NUM_LT = 7;
+
+  // Numerical 'less or equal than'
+  NUM_LE = 8;
 }
 
 // Short header fields
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Numeric operands for object search queries (#265)
 
 ### Changed
 
--- a/object/service.proto
+++ b/object/service.proto
@@ -496,6 +496,9 @@ message SearchRequest {
     // objects. Most human users expect to get only object they can directly
     // work with. In that case, `$Object:ROOT` filter should be used.
     //
+    // If `match_type` field is numerical, both `value` field and object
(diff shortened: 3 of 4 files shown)


roman-khimov (dismissed):

object/types.proto
```diff
@@ -47,6 +47,18 @@ enum MatchType {
...
+  NUM_GT = 5;
+
+  // Numerical 'greater or equal than'
+  NUM_GE = 6;
+
+  // Numerical 'less than'
+  NUM_LT = 7;
+
+  // Numerical 'less or equal than'
+  NUM_LE = 8;
```
carpawell, line 61: `NUM_EQUAL`?

> the previously added STRING_EQUAL / STRING_NOT_EQUAL matchers are suitable for numerical == / !=.

While it is true for now, it looks strange to me. Users should use `STRING_EQUAL` for numbers in their interfaces. Also, it is possible not to have strings-only values in the future.

carpawell (approved):

object/types.proto
cthulhu-rider: new values will be added when the existing ones are no longer sufficient

cthulhu-rider:
rebased and CHANGELOGed

object/types.proto
roman-khimov: I had this question as well, but as we're not supporting different notations ("1e8", for example) now, current set is sufficient for the purpose.

roman-khimov (approved):

object/types.proto
carpawell: my main point, i would say in the naming of this message: `MatchType` (it is not bound to `Filter`), no info about what it relates to. i can imagine reusing this for a number representation of smth, then we will have to add `NUM_EQUAL` but we will already have had some users that will be using `STRING_EQUAL`, strange

carpawell: > this enum is not intended to be reused

but there are no words about what it is used for

carpawell (approved):

Issue #279, opened 2023-12-01
Drop __NEOFS__TICK things


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I see unused magic attributes.

## Describe the solution you'd like
Drop __NEOFS_TICK from the specification. No one will ever use them, people never ever think of notifications in this way. We can try teaching them to do so, but likely what they want is for this to work without any additional per-object setups.

Can be deprecated as well if we care about v2.

## Describe alternatives you've considered
Keep maintaining it forever.

PR #287, opened 2024-02-22
object: Purge notification attributes

cthulhu-rider:


cthulhu-rider:
@roman-khimov deprecate means keep supporting, while lookin at #279 i thought we wanna stop supporting. CMIIW

roman-khimov:
There is a difference of API-level and node-level for us. I want to drop this from node, obviously, but API-wise we can still remember these were used (and not add any additional meanings to them, for example).

cthulhu-rider:
got it, done

carpawell (approved):

PR #288, opened 2024-02-22
session: Purge epoch headers

cthulhu-rider:


roman-khimov (commented):
Also deprecate, not drop? We are aiming to have stable API. Although the line is very blurry here, these are just attribute names.

cthulhu-rider:
as well as https://github.com/nspcc-dev/neofs-api/pull/287#issuecomment-1963815604

done

roman-khimov (dismissed):

roman-khimov (approved):

PR #294, opened 2024-03-05
object: Switch to object ID signature in `Replicate` request

cthulhu-rider:
* see also https://github.com/nspcc-dev/neofs-sdk-go/pull/565

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -717,8 +717,7 @@ message ReplicateRequest {
   // Object to be replicated.
   neo.fs.v2.object.Object object = 1;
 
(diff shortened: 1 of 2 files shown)


carpawell (approved):

carpawell:
BTW, not sure that is a backward change.

roman-khimov (approved):
We have a lot of problems with replays technically, not that this patch _adds_ it, but something that makes it more visible probably.

Regarding compatibility --- let's consider it to be a semantic fix, numbers are the same.

PR #301, opened 2024-07-05
Extend Target comment for ACL

### LINKED ISSUE #278: Account-based EACL targets
I'm always frustrated when I see public keys in the protocol used instead of accounts. Account is a very powerful abstraction that is not limited to a single key. Current EACL target requirement to have public keys leads to problems in the service layer like nspcc-dev/neofs-s3-gw#861 and while workarounds are possible the root cause of the issue is the requirement to have public keys in targets, they should not be needed to process EACL.

Extend target definition with a list of accounts. Deprecate key usage.

Reusing the same field is technically possible (it's just bytes). Not sure it's any better, but you can try convincing me.

smallhive:
Closes #278.

### DIFF
--- a/acl/types.proto
+++ b/acl/types.proto
@@ -176,7 +176,8 @@ message EACLRecord {
     // Target subject's role class
     Role role = 1 [json_name="role"];
 
-    // List of public keys to identify target subject
+    // List of 25-byte accounts to identify target subjects.
+    // 33-byte public keys are also supported, however, they are deprecated and script hashes should be derived from them.
     repeated bytes keys = 2 [json_name="keys"];
   }
   // List of target subjects to apply ACL rule to
--- a/proto-docs/acl.md
+++ b/proto-docs/acl.md
@@ -152,7 +152,7 @@ keys to match.
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
 | role | [Role](#neo.fs.v2.acl.Role) |  | Target subject's role class |
-| keys | [bytes](#bytes) | repeated | List of public keys to identify target subject |
+| keys | [bytes](#bytes) | repeated | List of 25-byte accounts to identify target subjects. 33-byte public keys are also supported, however, they are deprecated and script hashes should be derived from them. |
 
 
 <a name="neo.fs.v2.acl.EACLTable"></a>


acl/types.proto
```diff
@@ -176,7 +176,7 @@ message EACLRecord {
     // Target subject's role class
     Role role = 1 [json_name="role"];
 
-    // List of public keys to identify target subject
+    // List of public keys or accounts (scripthash from public key) to identify target subject
```
carpawell, line 179: have we decided if it is 25 or 20 bytes? tbh, i do not understand it from "scripthash from public key". also, i think it is critical to mention lengths now since it must define how to interpret this field

cthulhu-rider: `from public key` is definitely wrong in general (multisigs). I'd say
```suggestion
    // List of 20-byte account script hashes to identify target subjects. 33-byte public keys are also supported, however, they are deprecated and script hashes should be derived from them.
```

roman-khimov: We identify users with 25-byte value in NeoFS (see `OwnerID`). While it's a pure accidental choice and I really don't like it, these additional 5 bytes can be leveraged in various ways (for example, to support Bane accounts), so I propose using the same 25-byte IDs here. This would make it more consistent with the other parts of the protocol and also leave some space for future extensions.

roman-khimov (changes requested):

acl/types.proto
```diff
@@ -176,7 +176,8 @@ message EACLRecord {
     // Target subject's role class
     Role role = 1 [json_name="role"];
 
-    // List of public keys to identify target subject
+    // List of 25-byte accounts to identify target subjects.
+    // 33-byte public keys are also supported, however, they are deprecated and script hashes should be derived from them.
     repeated bytes keys = 2 [json_name="keys"];
```
cthulhu-rider, line 181: rename field to `accounts`?

cthulhu-rider (commented):
`make doc`

acl/types.proto
carpawell: btw, still not sure if we can do this. we even have `[json_name="keys"]` notation in our protocol

carpawell: (but i would just rename it)

roman-khimov (approved):

acl/types.proto
cthulhu-rider: i did not mean `json_name`. Protobuf field names are freely renamble. At the same time, divergence w/ json notation will come and this wouldnt be beautiful. So, since nobody cares much for now, lets keep

smallhive:
`make doc`'ed it

cthulhu-rider (approved):

roman-khimov (approved):

PR #303, opened 2024-07-11
refs: Prohibit IDs consisting of all zero bytes

cthulhu-rider:
i propose to reserve zero IDs and prohibit them from being message fields. While this is a breaking change in theory, in practice nothing will break. At the same time, application development will benefit

### DIFF
--- a/proto-docs/refs.md
+++ b/proto-docs/refs.md
@@ -71,7 +71,8 @@ content-addressed.
 
 `ContainerID` is a 32 byte long
 [SHA256](https://csrc.nist.gov/publications/detail/fips/180/4/final) hash of
-stable-marshalled container message.
+stable-marshalled container message. ID consisting of all zero bytes is
+reserved for undefined value and must not be specified as a field.
 
 String presentation is a
 [base58](https://tools.ietf.org/html/draft-msporny-base58-02) encoded string.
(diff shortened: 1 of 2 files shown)


carpawell (commented):
Overall agree. But do we really need to specify it for the fields that are commented as "hash"? What if container's hash is all zeros? It a caller is a liar (and that is not a 32 zeros hash), it is easy to check.

cthulhu-rider:
> What if container's hash is all zeros? It a caller is a liar (and that is not a 32 zeros hash), it is easy to check.

we see zeros - we dont hash the container at all. That's the difference

> But do we really need to specify it for the fields that are commented as "hash"?

for hash - mb not. But i prefer to explicitly state such cases. For example, although zeros are not a particularly valid hash, the protocol may dictate a special interpretation of them. And so we emphasize that it does not exist

roman-khimov (approved):

carpawell (approved):

roman-khimov:
Refs. https://datatracker.ietf.org/doc/html/rfc9562#name-nil-uuid

carpawell:
@roman-khimov, I have checked UUID generators _we used_ and it is impossible to get zero ID. At the same time, I do not understand what prevents getting zero sha256.

roman-khimov:
The probability of getting it is rather low. Hash collision between two containers (or objects) is more probable than that and if it happens -- we're toast.

PR #309, opened 2024-10-16
feat/add meta information consistency attribute

### LINKED ISSUE #300: Container consistency policy
I'm always frustrated when we can't specify consistency policy for a container.

Add an optional enum field with the following options:
 * available (just like the old one, no meta)
 * strict (nodes use meta and wait for meta acceptance)
 * optimistic (nodes use meta, but do not wait for acceptance)

If we want to alternate between different modes and preserve compatibility, there are not a lot of them.

carpawell:
Closes #300.

### DIFF
--- a/container/types.proto
+++ b/container/types.proto
@@ -52,6 +52,22 @@ message Container {
   //   accepted in a NeoFS network only if the global network hashing configuration
   //   value corresponds with that attribute's value. After container inclusion, network
   //   setting is ignored.
+  // * __NEOFS__METAINFO_CONSISTENCY \
+  //   Policy rule that defines the condition under which an object is considered
+  //   processed. Acceptable values and meanings:
+  //     1. "strict": SN processes objects' meta information, it is validated,
+  //       indexed and signed accordingly by a required minimal number of nodes
+  //       that are included in the container, a corresponding object inclusion
+  //       notification can be caught
+  //     2. "optimistic": the same as "strict" but a successful PUT operation
+  //       does not mean objects' meta information has been multi signed and
+  //       indexed correctly, however, SNs will try to do it asynchronously;
+  //       in general PUT operations are expected to be faster than in the
+  //       "strict" case
+  //     3. <other cases>: SN does not process objects' meta
+  //       information, it is not indexed and object presence/number of copies
+  //       is not proven after a successful object PUT operation; the behavior
+  //       is the same as it was before this attribute introduction
   //
   // And some well-known attributes used by applications only:
   //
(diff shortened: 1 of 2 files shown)


carpawell:
Naming is from the issue, have no suggestion from my side. I do not like that attribute idea, cause it is more like a core feature for me, every container gonna have it and it changes the game significantly, but agree that an attribute thing makes it more flexible.

carpawell:
Comments may be adjusted in the future when we will have more understanding of what the meta flow looks like.

container/types.proto
```diff
@@ -52,6 +52,22 @@ message Container {
   //   accepted in a NeoFS network only if the global network hashing configuration
   //   value corresponds with that attribute's value. After container inclusion, network
   //   setting is ignored.
+  // * __NEOFS__METAINFO_CONSISTENCY \
+  //   Policy rule that defines the condition under which an object is considered
+  //   processed. Acceptable values and meanings:
+  //     1. <empty>/<missing attribute>: SN does not process objects' meta
```
cthulhu-rider, line 58: attr value cannot be empty as stated above, absence is enough

container/types.proto
```diff
@@ -52,6 +52,22 @@ message Container {
...
+  //       information, it is not indexed and object presence/number of copies
+  //       is not proved after a successful object PUT operation; the behavior
+  //       is the same as it was before this attribute introduction
+  //     2. "strict": SN processes objects' meta information, it is validated,
+  //       indexed and signed accordingly by a required minimal number of nodes
+  //       that are included in the container, a corresponding object inclusion
+  //       notification can be catch
```
cthulhu-rider, line 65:
```suggestion
  //       notification can be caught
```

container/types.proto
```diff
@@ -52,6 +52,22 @@ message Container {
   //   accepted in a NeoFS network only if the global network hashing configuration
   //   value corresponds with that attribute's value. After container inclusion, network
   //   setting is ignored.
+  // * __NEOFS__METAINFO_CONSISTENCY \
+  //   Policy rule that defines the condition under which an object is considered
+  //   processed. Acceptable values and meanings:
+  //     1. <empty>/<missing attribute>: SN does not process objects' meta
+  //       information, it is not indexed and object presence/number of copies
+  //       is not proved after a successful object PUT operation; the behavior
```
roman-khimov, line 60: proven?

container/types.proto
carpawell: ok, reworded it to sound like "other cases" with the old behavior

container/types.proto
carpawell: afaik, both ways are ok currently, but don't mind

roman-khimov (approved):

PR #333, opened 2025-06-17
object: Allow header with payload chunk in first GET response message

cthulhu-rider:
here's an option through the proto version. It will work on the wire. The only other alternative I see is a new `GetV2` call which is more demanding of support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Payload chunk field to `object.PutRequest.Init` message (#333)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cthulhu-rider:
https://github.com/nspcc-dev/neofs-node/issues/3396#issuecomment-2980001356

carpawell (commented):
Do we have any load updates on it?

roman-khimov:
All of the old data suggests that it's irrelevant. But it should be tried again, 4K perf has changed seriously and we also have EC now that is different in some cases.

carpawell:
I dont like the idea of such changes in API, but it is faster, we should consider it.

roman-khimov:
The API change can be made compatible by introducing another type for `oneof`.

cthulhu-rider:
since new field is needed anyway, decided to just extend heading message. Thus, initial message can carry entire object completely

cthulhu-rider:
performance tests showed no impact

PR #336, opened 2025-06-25
object: deprecate ShortHeader

carpawell:
Its usage is unknown, current official implementations do not support and do not expect it.


carpawell:
I do not know if it is possible to add doc links in proto comments, likely not. Never tried and thought about it.

cthulhu-rider (approved):

PR #337, opened 2025-06-26
object: add `__NEOFS__ASSOCIATE` object attribute

### LINKED ISSUE #311: Limit TOMBSTONE members number to one
I'm always frustrated when there is only a single object in the [DELETE](https://github.com/nspcc-dev/neofs-api/blob/e66b25d4bf2afc4472023e1e2c2467f694f5a0e1/object/service.proto#L375) RPC but a lot of objects are in the tombstone's [body](https://github.com/nspcc-dev/neofs-api/blob/e66b25d4bf2afc4472023e1e2c2467f694f5a0e1/tombstone/types.proto#L25-L29). It is even more unexpected after https://github.com/nspcc-dev/neofs-api/pull/297 (however, it is not that clear when we are cleaning unfinished big objects).

Disallow "big" tombstones. One TS -- one object. Tombstones expire sooner or later and tombstones with many objects are not expected, it is even not possible currently due to DELETE RPC, only if someone creates it manually. 

Keep it as is.

## Additional context
Uncontrollable number of deleted objects by a single object makes it harder to maintain object meta information: https://github.com/nspcc-dev/neofs-contract/pull/448#discussion_r1863821697.

### LINKED ISSUE #312: Limit LOCK members number to one
Same as in #311 but for LOCK objects. It is a little bit harder because there is no LOCK RPC and every LOCK object is created from scratch on the [client side](https://github.com/nspcc-dev/neofs-node/blob/345f2317911eec9a9dd33c2b81993c4cdb5626b3/cmd/neofs-cli/modules/object/lock.go#L56-L57) currently.

It can also be restricted to one LOCK -- one LOCKed object. This also makes it more flexible for expiration and other side effects while LOCKing.

Keep it as is.

## Additional context
Uncontrollable number of locked objects by a single object makes it harder to maintain object meta information: https://github.com/nspcc-dev/neofs-contract/pull/448#discussion_r1863821697.

carpawell:
It obsoletes old LOCK- and TOMBSTONE-typed objects with strictly typed messages in their payloads. Closes #312, closes #311.

### DIFF
--- a/proto-docs/object.md
+++ b/proto-docs/object.md
@@ -929,6 +929,12 @@ that affect system behaviour:
 * __NEOFS__EXPIRATION_EPOCH \
   Tells GC to delete object after that epoch (but object is available
   throughout the epoch specified in this attribute).
+* __NEOFS__ASSOCIATE \
+  Associated object. For TOMBSTONE, LOCK object types it defines object
+  to delete and to lock accordingly. For objects of 2.18+ API version, it
+  is the only way to delete/lock objects. It MUST be a single stringified
+  (according to [refs.ObjectID] message) object ID with no leading or
+  trailing spaces.
 * __NEOFS__TICK_EPOCH \
   Decimal number that defines what epoch must produce
   object notification with UTF-8 object address in a
--- a/lock/types.proto
+++ b/lock/types.proto
@@ -7,6 +7,10 @@ import "refs/types.proto";
 option csharp_namespace = "Neo.FileStorage.API.Lock";
 option go_package = "github.com/nspcc-dev/neofs-sdk-go/proto/lock";
 
+// DEPRECATED: lock objects 1-to-1 with __NEOFS__ASSOCIATE object attribute in a
+// LOCK-typed object with no payload. For objects of 2.18+ API version, it is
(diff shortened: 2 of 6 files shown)


proto-docs/object.md
```diff
@@ -1044,6 +1047,8 @@ prefix to the name. Here is the list of fields available via this prefix:
   16 byte UUIDv4 used to identify the split object hierarchy parts
 * $Object:split.first \
   object_id of the first part in split chain; non-acceptable for deprecated V1 split scheme
+* $Object:target \
```
cthulhu-rider, line 1050: with `__NEOFS__TARGET` attribute this wouldnt've been needed. Any reason to not do it?

carpawell: this is a core feature for me with a certain type and non-string original representation. i feel it the same way as oid or other fields. likely all the necessary fields can be placed in the attributes in their string form, and SN may be forced to validate them, but what for?

carpawell: the same way, locking many objects with a single LOCK object will not be acceptable. non-compatible either

carpawell: @roman-khimov, thoughts?

PR #343, opened 2025-09-03
status: add PUT deny status on exceeding limits

carpawell:
Refs https://github.com/nspcc-dev/neofs-node/issues/3520.

### DIFF
--- a/status/types.proto
+++ b/status/types.proto
@@ -124,6 +124,10 @@ enum Object {
 
   // [**2053**] Invalid range has been requested for an object.
   OUT_OF_RANGE = 5;
+
+  // [**2054**] Quota was exceeded; no further PUTs will be allowed within
+  // current settings.
+  QUOTA_EXCEEDED = 6;
 }
 
 // Section of statuses for container-related operations.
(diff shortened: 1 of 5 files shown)


status/types.proto
```diff
@@ -124,6 +124,9 @@ enum Object {
 
   // [**2053**] Invalid range has been requested for an object.
   OUT_OF_RANGE = 5;
+
+  // [**2054**] Size limit was exceeded, no further PUTs will be allowed.
+  SIZE_LIMIT_EXCEEDED = 6;
```
roman-khimov, line 129: QUOTA_EXCEEDED?

cthulhu-rider: it's general purpose to me, so i'd make it a `RESOURCE_EXHAUSTED` in `CommonFail`. By analogy with [gRPC](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) or kinda [HTTP](https://datatracker.ietf.org/doc/html/rfc6585#section-4)

roman-khimov: > RESOURCE_EXHAUSTED in CommonFail

BTW, HTTP example is a temporary error, retrying the same request in 10 minutes can succeed. Quotas are different, this error is permanent unless either quota is changed or some other objects are deleted.

PR #345, opened 2025-09-08
Support EC in container storage policies

cthulhu-rider:


object/service.proto
```diff
@@ -262,7 +262,17 @@ service ObjectService {
...
 }
 
-// GET object request
+// GET object request.
+//
+// The query for a parent object's EC part locally stored on the server is
+// specified as follows:
+//  - `body.address` is an address of the parent;
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
+//    `__NEOFS__EC_PART_IDX` by object attribute format.  Rule index MUST NOT
```
carpawell, line 271:
```suggestion
//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
```

object/service.proto
```diff
@@ -653,7 +663,17 @@ message Range {
...
 }
 
-// Request part of object's payload
+// Request part of object's payload.
+//
+// The query for a parent object's EC part locally stored on the server is
+// specified as follows:
+//  - `body.address` is an address of the parent;
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
+//    `__NEOFS__EC_PART_IDX` by object attribute format.  Rule index MUST NOT
```
carpawell, line 672:
```suggestion
//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
```

netmap/types.proto
```diff
@@ -130,6 +130,53 @@ message PlacementPolicy {
     json_name = "subnetId",
     deprecated = true
   ];
+
+  // Erasure coding rule for container objects.
+  //
+  // For each original object, the payload is split into `data_part_num` data
+  // and `parity_part_num` parity parts. Each part is the same size. Data parts
+  // contain the original payload. If its length is not divisible by
```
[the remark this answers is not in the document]
cthulhu-rider: it is correct

EESergey (approved):

netmap/types.proto
```diff
@@ -130,6 +130,53 @@ message PlacementPolicy {
...
+    // Number of data parts
+    uint32 data_part_num = 1;
+
+    // Number of parity parts
+    uint32 parity_part_num = 2;
+
+    // Name of the linked selector
+    string selector = 3 [json_name = "selector"];
+  }
+  // Erasure coding rules. Limited to 256 items.
```
roman-khimov, line 178: Limit to 1? Yeah, I know it's an array and theoretically many things are possible. But I'm not sure what will happen with "EC 3/1 EC 12/4 EC 5/3 EC 4/2" policy. Maybe it'll work, but I doubt it'll be usable. Also it's easier to extend things in future rather than constrain. So maybe 2-4, but no more.

netmap/types.proto
```diff
@@ -130,6 +130,53 @@ message PlacementPolicy {
...
+  repeated ECRule ec_rules = 6 [json_name = "ecRules"];
```
roman-khimov, line 179: Also it's completely separated from `replicas` which makes these rules kinda alternative. It can be an implementation detail, but the fact that they're not on the same level can be confusing.

object/types.proto
```diff
@@ -228,6 +228,12 @@ message Header {
   // * __NEOFS__TICK_TOPIC \
   //   UTF-8 string topic ID that is used for object notification.
   //   DEPRECATED: attribute ignored by servers.
+  // * __NEOFS__EC_RULE_IDX \
+  //   Index of EC rule in container's `PlacementPolicy.ec_rules` according to
```
roman-khimov, line 232: Like this index. For me in "REP 3 EC 3/1" the EC rule index is 1. But in fact it's 0.

object/service.proto
```diff
@@ -262,7 +262,17 @@ service ObjectService {
   rpc Replicate(ReplicateRequest) returns (ReplicateResponse);
 }
 
-// GET object request
+// GET object request.
+//
+// The query for a parent object's EC part locally stored on the server is
```
roman-khimov, line 267: But it can be retrieved via part OID as well, right?

object/service.proto
```diff
@@ -262,7 +262,17 @@ service ObjectService {
...
+// specified as follows:
+//  - `body.address` is an address of the parent;
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
+//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
```
roman-khimov, line 271: What if netmap change happens and we're requesting part 0/4 from node that stores 0/1?

object/service.proto
cthulhu-rider: yes, w/ regular GET request

netmap/types.proto
cthulhu-rider: i doubt various encoding parameters will be in demand too. But multiple settings may be required to bind to different selectors. For example, they can define disjoint subnets for placement

picked 256 from `replicas`

netmap/types.proto
cthulhu-rider: they are at the same level of policy, just diff fields. I though that we can add `parity` field to `Replica`. Zero stands for old REP rule, non-zero - for EC one. But the storage scheme is so different, so i decided not to merge them. And IINM there wont be any gain in size

netmap/types.proto
roman-khimov: Let's bring it down to four.

netmap/types.proto
roman-khimov: As discussed yesterday policy string representation is already a mix of different things from the same level (rep/cbf/selectors) and it has no order logically. Since EC is a new features that differs significantly in behavior from replicas having another field in protobuf is more convenient in many cases.

PR #346, opened 2025-09-23
New status codes

roman-khimov:


status/types.proto
```diff
@@ -81,6 +81,12 @@ enum Success {
   // If the server cannot match successful outcome to the code, it should
   // use this code.
   OK = 0;
+
+  // [**1**] Incomplete success, only a part of the request done which may
+  // or may not be sufficient for the client. Examples are object PUT making
+  // 2 out of 3 required replicas or distributed SEARCH getting result from
+  // 4 out of 5 nodes.
+  INCOMPLETE = 1;
```
cthulhu-rider, line 89: this should also be mentioned in https://github.com/nspcc-dev/neofs-api/blob/774e6bb7e204b679a24423fee117695341978111/object/service.proto#L64 for PUT and SEARCH

cthulhu-rider (commented):
useful ones

status/types.proto
roman-khimov: Fixed.

cthulhu-rider (approved):

Issue #241, opened 2022-09-05
Bearer token in request classification


alexvanin:
# Description 

We don't have detailed description how NeoFS Node classifies requests to apply **basic** and **extended** ACL rules.

Currently NeoFS Node fetches the key and compares it with container owner, alphabet or container nodes. Here is the simplified scheme of fetch function from latest release v0.31.0: [RequestOwner()](https://github.com/nspcc-dev/neofs-node/blob/cd71de69a08d87a63fedd7bc4ead54c4ddac8791/pkg/services/object/acl/v2/request.go#L111-L137).
```mermaid
flowchart LR
    A[Start] --> B{Is session\ntoken present?}
    B -->|Yes| C[Owner == Session token creator]
    B -->|No| D[Owner == Request sender]
    C --> E[End]
    D --> E
```

The only way to impersonate user is to use session token signed by user's key. The issue is that session token can't be reused and usually is signed **on every request**: object context requires [verb](https://github.com/nspcc-dev/neofs-api/blob/15eaf93ba1eb6947fb6e0927a1e6c430866cfa52/session/types.proto#L41) and [object address](https://github.com/nspcc-dev/neofs-api/blob/15eaf93ba1eb6947fb6e0927a1e6c430866cfa52/session/types.proto#L44) which are frankly updated on every request.

Session token looks like a nice solution for peer-to-peer communication where client side can use private key and sign every request. However it might not be a best case when gateway appears between user's client and neofs-node.
1) Signing every NeoFS request might be very bad UX (e.g. with Wallet Connect)
2) Session token is not transitive, so the client has to create session with unknown NeoFS Node, instead of the known gateway.

For gateways we usually work with bearer token on client side to control object access, while session is established between gateway and the node, as it was originally designed.

```mermaid
flowchart LR
    A[Panel.NeoFS] o--o|Bearer token| B[REST Gateway]
    B o--o|Session token| C[NeoFS Node]
    B o--o|Bearer token| C[NeoFS Node]
```

:green_circle: Nicer UX where user signs tokens once
:red_circle: It works only with **public containers**, because gateway requests are always classified as `OTHERS` group. 

Basic ACL check will DENY such requests on private container. Therefore private container in fact should be public with restrictive eACL (this is how S3 Gateway works with buckets).

# Proposal

Bearer token structure can be expanded with a bool flag to allow request impersonation.

```diff
message BearerToken {
  message Body {
[...]

PR #350, opened 2025-10-20
session: add a new session token v2

End-rey:
Refs #241.

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -9,6 +9,48 @@ import "status/types.proto";
 option csharp_namespace = "Neo.FileStorage.API.Session";
 option go_package = "github.com/nspcc-dev/neofs-sdk-go/proto/session";
 
+// Verb represents all possible operations in NeoFS that can be authorized
+// via session tokens or delegation chains. This enum covers both object and
+// container service operations.
+enum Verb {
+  // Unknown verb
+  VERB_UNSPECIFIED = 0;
+
+  // Object operations
+
+  // Refers to object.Put RPC call
+  OBJECT_PUT = 1;
+
+  // Refers to object.Get RPC call
+  OBJECT_GET = 2;
+
+  // Refers to object.Head RPC call
+  OBJECT_HEAD = 3;
+
+  // Refers to object.Search RPC call
+  OBJECT_SEARCH = 4;
+
+  // Refers to object.Delete RPC call
+  OBJECT_DELETE = 5;
+
+  // Refers to object.GetRange RPC call
+  OBJECT_RANGE = 6;
+
+  // Refers to object.GetRangeHash RPC call
+  OBJECT_RANGEHASH = 7;
+
+  // Container operations
+
+  // Refers to container.Put RPC call
+  CONTAINER_PUT = 8;
+
+  // Refers to container.Delete RPC call
+  CONTAINER_DELETE = 9;
+
+  // Refers to container.SetExtendedACL RPC call
+  CONTAINER_SETEACL = 10;
+}
+
 // Context information for Session Tokens related to ObjectService requests
 message ObjectSessionContext {
   // Object request verbs
@@ -175,6 +217,10 @@ message RequestMetaHeader {
   // Session token within which the request is sent
   SessionToken session_token = 5 [json_name = "sessionToken"];
 
+  // Session token v2 with delegation chain support.
+  // Requests are invalid if both session_token and session_token_v2 are set.
+  SessionTokenV2 session_token_v2 = 9 [json_name = "sessionTokenV2"];
+
   // `BearerToken` with eACL overrides for the request
   neo.fs.v2.acl.BearerToken bearer_token = 6 [json_name = "bearerToken"];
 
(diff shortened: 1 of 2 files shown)


session/types.proto
```diff
@@ -232,3 +233,153 @@ message ResponseVerificationHeader {
   // Chain of previous hops signatures
   ResponseVerificationHeader origin = 4 [json_name = "origin"];
 }
+
+// Session Token v2
+
+// Account represents an identity in NeoFS.
+// It can be either direct (OwnerID) or indirect (NNS domain).
+message Account {
```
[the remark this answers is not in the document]
carpawell: it would look better there for me, yes

refs/types.proto
```diff
@@ -79,6 +79,19 @@ message OwnerID {
...
 
+// Account represents an identity in NeoFS.
+// It can be either direct (OwnerID) or indirect (NNS domain).
+message Account {
+  // Account identifier.
+  oneof identifier {
+    // Direct account reference via OwnerID (hash of public key).
+    OwnerID owner_id = 1 [json_name = "ownerID"];
+
+    // Indirect account reference via NeoFS Name Service.
```
roman-khimov: What's the expected NNS contents? This should be described as well.

session/types.proto
```diff
@@ -232,3 +233,140 @@ message ResponseVerificationHeader {
...
+    // Lifetime of this token.
+    TokenLifetime lifetime = 5 [json_name = "lifetime"];
+
+    // Authorization context - what operations are allowed.
+    oneof context {
+      // ObjectService authorization context.
+      ObjectSessionContextV2 object = 6 [json_name = "object"];
+
+      // ContainerService authorization context.
+      ContainerSessionContextV2 container = 7 [json_name = "container"];
```
roman-khimov: Maybe we can have a single set of object/container operations? I'm not sure what this separation buys us.

session/types.proto
```diff
@@ -85,6 +127,21 @@ message ContainerSessionContext {
   refs.ContainerID container_id = 3 [json_name = "containerID"];
 }
 
+// Lifetime parameters of the token. Field names taken from rfc7519.
+message TokenLifetime {
+  // Expiration epoch, the last epoch when token is valid.
+  // For SessionTokenV2 this is the last valid Unix timestamp.
```
roman-khimov: 1. It's very different in semantics.
2. Compatibility warning fires for the removal of old `TokenLifetime`, so we can keep it as is and add this new one for v2.

session/types.proto
```diff
@@ -232,3 +282,96 @@ message ResponseVerificationHeader {
...
+  oneof identifier {
+    // Direct account reference via OwnerID (hash of verification script).
+    neo.fs.v2.refs.OwnerID owner_id = 1 [json_name = "ownerID"];
+
+    // Indirect account reference via NeoFS Name Service.
+    // NNS name is a domain name that resolves to an OwnerID through the
+    // NeoFS Name Service. The name must be a valid DNS-like domain name
+    // (e.g., "example.neofs") that is registered in the NNS contract on
+    // the Neo blockchain. The NNS record should contain a string record with
+    // the corresponding OwnerID value.
```
roman-khimov: Mention NEP-18?

session/types.proto
```diff
@@ -232,3 +282,96 @@ message ResponseVerificationHeader {
...
+    // Token version.
+    uint32 version = 1 [json_name = "version"];
+
+    // Token identifier (UUIDv4 in binary form).
+    bytes id = 2 [json_name = "id"];
+
+    // Account that issued this token (who signed it).
+    Target issuer = 3 [json_name = "issuer"];
+
+    // Accounts authorized by this token (who can use it).
```
roman-khimov: We need some limits for all repeated fields.

End-rey: Now I have added comments. However, I see that they can be validated at runtime with https://github.com/bufbuild/protoc-gen-validate.
And also, some limits are discussable.

session/types.proto
```diff
@@ -232,3 +278,114 @@ message ResponseVerificationHeader {
...
+    // Maximum number of subjects: 100.
```
roman-khimov: 8 for now, given that we can have proper groups this shouldn't be used for hundreds of things.

session/types.proto
```diff
@@ -232,3 +278,114 @@ message ResponseVerificationHeader {
...
+  // Accounts that received the delegation.
+  // Maximum number of subjects: 100.
+  repeated Target subjects = 2 [json_name = "subjects"];
+
+  // Lifetime of this delegation.
+  // Allows each delegation in the chain to have its own validity window.
+  TokenLifetime lifetime = 3 [json_name = "lifetime"];
+
+  // List of verbs authorized by this delegation.
+  repeated Verb verbs = 4 [json_name = "verbs"];
```
roman-khimov: Why verbs here, btw? Seem to be duplicating context data.

session/types.proto
```diff
@@ -232,3 +278,114 @@ message ResponseVerificationHeader {
...
+  // Not valid before, the first valid Unix timestamp.
+  uint64 nbf = 2 [json_name = "nbf"];
+
+  // Issued at, the Unix timestamp when the token was issued.
+  uint64 iat = 3 [json_name = "iat"];
+}
+
+
+// DelegationInfo represents a single delegation in a chain of trust.
+message DelegationInfo {
```
roman-khimov: We need some "final" bit as well for the case where token can't be used to build longer chains (A trusts B to do something, but doesn't give B permission to issue tokens for C).

session/types.proto
End-rey: I used them here with the idea that not all verbs can be delegated.

session/types.proto
```diff
@@ -232,3 +278,97 @@ message ResponseVerificationHeader {
...
+
+// SessionTokenV2 represents NeoFS Session Token with delegation support.
+message SessionTokenV2 {
+  // Session Token body.
+  message Body {
+    // Token version.
+    uint32 version = 1 [json_name = "version"];
+
+    // Token identifier (UUIDv4 in binary form).
+    bytes id = 2 [json_name = "id"];
```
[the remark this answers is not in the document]
cthulhu-rider: does anyone see usecase for creating additional token that differs only in `id`?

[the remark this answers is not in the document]
roman-khimov: I treat it like a transaction nonce. Same version, same policy, same everything, but we want to have a different ID of course. For that UUID is an overkill.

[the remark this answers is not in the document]
roman-khimov: But why would anyone want to create the same token with different ID is a mystery. Maybe it can happen accidentally though. Like some code has lost the old token and then it creates a new one.

carpawell (approved):

session/types.proto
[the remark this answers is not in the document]
cthulhu-rider: currently, sys behavior is identical for two tokens that differ only in `id`. For example, the fact that some token was lost and recreated on user side (this is possible with RFC6979 signature scheme) still cannot be verified. However, caching will require more resources

session/types.proto
```diff
@@ -175,6 +217,10 @@ message RequestMetaHeader {
   // Session token within which the request is sent
   SessionToken session_token = 5 [json_name = "sessionToken"];
 
+  // Session token v2 with delegation chain support.
+  // If both session_token and session_token_v2 are set, session_token_v2 takes precedence.
```
cthulhu-rider: i'd consider such requests invalid

session/types.proto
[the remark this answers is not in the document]
End-rey: So, drop `id` and use for caching hash body?

session/types.proto
End-rey: We once discussed that both tokens are possible and they do not interfere with each other @roman-khimov

session/types.proto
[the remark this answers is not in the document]
roman-khimov: Replace is with 32 bit nonce. Just in case we need some randomness. Otherwise they're identified by hashes (like containers or objects are, for example).

session/types.proto
```diff
@@ -232,3 +278,98 @@ message ResponseVerificationHeader {
...
+}
+
+// SessionTokenV2 represents NeoFS Session Token with delegation support.
+message SessionTokenV2 {
+  // Session Token body.
+  message Body {
+    // Token version.
+    uint32 version = 1 [json_name = "version"];
+
+    // Unique token identifier.
```
roman-khimov: It's not unique, it's just a number to prevent collisions.

cthulhu-rider (approved):

PR #369, opened 2025-12-24
Container+EACL

cthulhu-rider:


carpawell:
I have a PoC with the updated version, but i'm still not sure whether we need to complicate session things and have an additional signature field. They are used to ensure that setting eACL is an authorized operation, but if a user is already able to create a container, why not just allow him to change eACL in the same operation? I would treat eACL as a container attribute/field, this would simplify API and server validation. I cannot imagine cases when we have some policies that must allow container creation, but not changing its eACL, even with N3 signatures and complex validation logic.

roman-khimov (approved):
I think this structure is flexible enough to support different implementations. Ideally signature check should be performed at the contract level, in which case a separate signature for a separate call makes total sense.

roman-khimov (approved):

PR #389, opened 2026-04-22
Deprecate rangehash, replace range with parameterized get

roman-khimov:


object/service.proto
```diff
@@ -282,6 +286,13 @@ message GetRequest {
     // If `raw` flag is set, request will work only with objects that are
     // physically stored on the peer node
     bool raw = 2;
+
+    // Requested payload range (whole payload if not specified).
+    Range range = 3;
+
+    // If set, makes Get return payload only, completely omitting Init response
+    // message with header data.
+    bool payload_only = 4;
```
roman-khimov, line 295: Notice that this is also important wrt EACL security. Currently forwarding nodes can't check RANGE requests wrt EACLs containing object fields.

carpawell: and what do we plan to do with it? forward always with headers to check eACL, but return payload only to the client?

carpawell (approved):

object/service.proto
roman-khimov: In case EACL requires it.

PR #393, opened 2026-06-22
object/GET: allow to fetch exact EC part in GET request

### LINKED ISSUE #348: Allow to get EC chunk by rule idx only
See https://github.com/nspcc-dev/neofs-api/pull/345#discussion_r2401240576, rule/part is good for policer, but suboptimal for regular GET in case there was some change in netmap and part are placed suboptimally. Unlike with REP where a shift by 1 in node list is unnoticeable in most cases, with EC this means 12 failures for 12 tries to get a part in 12+4 scenario. The proposal is to allow to GET whatever part node has by rule only and leverage that in node. Additional testing can be helpful, but protocol-wise it's a rather simple thing.

carpawell:
Closes #348.

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -271,10 +271,22 @@ service ObjectService {
 // The query for a parent object's EC part locally stored on the server is
 // specified as follows:
 //  - `body.address` is an address of the parent;
-//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
-//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
-//    exceed container's `PlacementPolicy.ec_rules` list. Part index MUST NOT
-//    exceed total part number in the indexed rule.
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` by object
+//    attribute format. Rule index MUST NOT exceed container's
+//    `PlacementPolicy.ec_rules` list.
+//    If `__NEOFS__EC_PART_IDX` is also included in X-headers, node returns
+//    corresponding part. Part index MUST NOT exceed total part number in the
+//    indexed rule. If index is unspecified:
+//    - if there is a single part in storage node: it is returned;
+//    - if there is more than a single part: the appropriate part for the
+//      current storage node (according to the actual Network map) is returned;
+//      if storage node does not have the appropriate part, the lowest index
+//      part it has is returned;
+//    - if `body.payload_only` is set to `true`, response's
+//      `meta_header.x_headers` MUST include `__NEOFS__EC_PART_IDX` in the same
+//      form to identify returned object part;
+//    - if `body.range` is specified, error is returned, meaning any ranged GET
+//      MUST have `__NEOFS__EC_PART_IDX` set.
 // In this case, if `body.address` refers to TOMBSTONE or LOCK object (which
 // cannot have EC parts), the query applies to it.
 message GetRequest {
--- a/proto-docs/object.md
+++ b/proto-docs/object.md
@@ -556,10 +556,22 @@ GET object request.
 The query for a parent object's EC part locally stored on the server is
 specified as follows:
  - `body.address` is an address of the parent;
- - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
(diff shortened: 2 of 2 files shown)


object/service.proto
```diff
@@ -271,10 +271,15 @@ service ObjectService {
...
-//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
-//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
-//    exceed container's `PlacementPolicy.ec_rules` list. Part index MUST NOT
-//    exceed total part number in the indexed rule.
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` by object
+//    attribute format. Rule index MUST NOT exceed container's
+//    `PlacementPolicy.ec_rules` list.
+//    If `__NEOFS__EC_PART_IDX` is also included in X-headers, node returns
+//    corresponding part, otherwise nodes returns any part for the rule
+//    described above it has (if more that one part were found, the order
```
carpawell: i find it dangerous when netmap is changing, and there could be policer's object changes. the same is about any form of GETRANGE

object/service.proto
```diff
@@ -271,10 +271,15 @@ service ObjectService {
 // The query for a parent object's EC part locally stored on the server is
 // specified as follows:
 //  - `body.address` is an address of the parent;
-//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` and
-//    `__NEOFS__EC_PART_IDX` by object attribute format. Rule index MUST NOT
-//    exceed container's `PlacementPolicy.ec_rules` list. Part index MUST NOT
-//    exceed total part number in the indexed rule.
+//  - `meta_header.x_headers` includes `__NEOFS__EC_RULE_IDX` by object
```
cthulhu-rider, line 274: why not do the same for `GetRange`?

object/service.proto
cthulhu-rider: > order is undefined

since parts are streamed to the client one-by-one, i think it is worth guaranteeing direct order. Is there any reason not to do so?

carpawell: do you suggest returning every part in the same GET request in case of any conflicts?

object/service.proto
carpawell: as i stated in the thread above: i am afraid that we will have many problems with undefined requests (no part index x-headers) returning objects if no header is received in the same request (any GETRANGE/GET with range and `payload_only`) is not safe cause it is impossible to know what exact object was handled

object/service.proto
roman-khimov: Multipart can be solved by returning "appropriate" (as per current placement) or lower index (if there is no appropriate) part.

object/service.proto
```diff
@@ -271,10 +271,15 @@ service ObjectService {
...
+//    attribute format. Rule index MUST NOT exceed container's
+//    `PlacementPolicy.ec_rules` list.
+//    If `__NEOFS__EC_PART_IDX` is also included in X-headers, node returns
+//    corresponding part, otherwise nodes returns any part for the rule
+//    described above it has (if more that one part were found, the order
+//    is undefined). Part index MUST NOT exceed total part number in the
+//    indexed rule. `__NEOFS__EC_PART_IDX` MUST NOT be omitted with
+//    non-nil `body.range` or `body.payload_only` field set to `true`.
```
roman-khimov: Range does require idx for sure, but payload only can be solved with x header _returned_ from node.

roman-khimov (commented):
The spec is rather simple, however this raises questions over forwarding optimizations, these require strict ordering and we can end up with more simultaneous connections.

object/service.proto
carpawell: > Multipart can be solved by returning "appropriate"

from which side? client or server? what if they are different? if we believe that netmap is ok (not changing right now) why is this optimisation needed at all?

object/service.proto
carpawell: i vote for returning this header _always_ then

roman-khimov: An option too.

object/service.proto
roman-khimov: Only server (the one processing the request) can decide here.

roman-khimov (approved):

PR #399, opened 2026-07-10
object: add flexible payload range forms

### LINKED ISSUE #395: HTTP-alike bytes range specification
NeoFS uses `offset:size` for ranged GET. REST/S3 gateways often accept HTTP range specifications for bytes. In simple cases like `A-B` it can be converted into NeoFS one, but in this case NeoFS doesn't follow this rule:

> If the [last-pos](https://www.rfc-editor.org/info/rfc9110/#rule.int-range) value is absent, or if the value is greater than or equal to the current length of the representation data, the byte range is interpreted as the remainder of the representation (i.e., the server replaces the value of [last-pos](https://www.rfc-editor.org/info/rfc9110/#rule.int-range) with a value that is one less than the current length of the selected representation).

And in many cases it can't be converted at all, `A-` or `-X` all require prior knowledge of object length that gateways obtain via HEAD request. This makes gateways less efficient for ranged requests and real applications dealing with various data formats can issue a lot of ranged requests.

Add support for HTTP-alike ranged GET semantics.

## Additional context
https://www.rfc-editor.org/info/rfc9110/#name-byte-ranges

End-rey:
Closes #395.

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -305,6 +305,9 @@ message GetRequest {
     // If set, makes Get return payload only, completely omitting Init response
     // message with header data.
     bool payload_only = 4;
+
+    // Requested extended payload range. MUST NOT be set together with `range`.
+    ExtendedRange extended_range = 5;
   }
   // Body of get object request message.
   Body body = 1;
@@ -704,6 +707,17 @@ message Range {
 
   // Length in bytes of the object payload range
   uint64 length = 2;
+}
+
+// Extended object payload range.
+message ExtendedRange {
+  // Offset of the first byte from the object payload start. If omitted, the
+  // range is defined by `last_pos` as the number of bytes from the payload end.
+  optional uint64 first_pos = 1;
+
+  // Offset of the last byte from the object payload start. If omitted, the
+  // range starts at `first_pos` and extends to the end of the payload.
+  optional uint64 last_pos = 2;
 }
 
 // Request part of object's payload.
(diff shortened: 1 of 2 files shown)


object/service.proto
```diff
@@ -695,15 +695,37 @@ message SearchV2Response {
...
 }
 
-// Object payload range. Ranges of zero length SHOULD be considered as invalid
-// except for the special 0:0 request which is interpreted as "get whole
-// payload" and allows to receive payload only when its size is not known.
+// Object payload range. By default range is interpreted as an (offset, length)
+// tuple. Ranges of zero length SHOULD be considered as invalid except for the
+// special 0:0 request which is interpreted as "get whole payload" and allows to
+// receive payload only when its size is not known.
 message Range {
```
roman-khimov, line 704: It's used for RANGE requests as well, so kinda implies that it should be extended too. Practically can be ignored though.

My first idea was to extend GetRequest with an alternative definition of range, can be made like `ExtendedRange` or something.

object/service.proto
```diff
@@ -695,15 +695,37 @@ message SearchV2Response {
...
+  }
+
   // Offset of the range from the object payload start
   uint64 offset = 1;
 
   // Length in bytes of the object payload range
   uint64 length = 2;
+
+  // Alternative range forms resolved against the object payload size.
+  oneof kind {
```
roman-khimov: Can be just two numbers as well. Likely would be easier to handle in code.

object/service.proto
carpawell: i like the idea of extending only GET's range. maybe it would be even possible to do it without breaking changes

End-rey: Done this way.

object/service.proto
End-rey: Done.

cthulhu-rider (approved):

roman-khimov (approved):
Nice use of `optional`.

carpawell (approved):

PR #401, opened 2026-07-20
acl: add comment field to EACL records

### LINKED ISSUE #400: EACL comment field
I'm always frustrated when we can't put any additional per-record data into EACL. Some applications might want to store meta associated with rule, users may want to have comments for rules.

Add "comment" field to EACL record that is not affecting any permission logic, but allows to store auxiliary data if needed. Make it UTF-8.

Not a lot of them.

End-rey:
Closes #400.

### DIFF
--- a/acl/types.proto
+++ b/acl/types.proto
@@ -182,6 +182,10 @@ message EACLRecord {
   }
   // List of target subjects to apply ACL rule to
   repeated Target targets = 4 [json_name = "targets"];
+
+  // Optional auxiliary data associated with the rule. The comment does not
+  // affect permission evaluation and must be valid UTF-8 without zero bytes.
+  string comment = 5 [json_name = "comment"];
 }
 
 // Extended ACL rules table. A list of ACL rules defined additionally to Basic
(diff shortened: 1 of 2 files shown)


acl/types.proto
```diff
@@ -182,6 +182,10 @@ message EACLRecord {
   }
   // List of target subjects to apply ACL rule to
   repeated Target targets = 4 [json_name = "targets"];
+
+  // Optional auxiliary data associated with the rule. The comment does not
+  // affect permission evaluation and must be valid UTF-8.
```
roman-khimov: ... without zero bytes

Just to remind, 0x00 is a valid UTF-8 character, but we never want that.

End-rey: Done.

acl/types.proto
```diff
@@ -182,6 +182,10 @@ message EACLRecord {
   }
   // List of target subjects to apply ACL rule to
   repeated Target targets = 4 [json_name = "targets"];
+
+  // Optional auxiliary data associated with the rule. The comment does not
+  // affect permission evaluation and must be valid UTF-8 without zero bytes.
```
cthulhu-rider, line 187: i'd also limit its len. Start with a conservative value like 1KiB

roman-khimov: That's somewhat more complicated, related to limits for targets and keys. Looks like we don't have any in API now. At the same time we have a hard overall limit for EACL (#398).

roman-khimov (approved):

cthulhu-rider (approved):

PR #402, opened 2026-07-21
session: deprecate origin meta headers and signature

### LINKED ISSUE #387: Drop matryoshka from meta headers
I'm always frustrated when I see this:
```
// Meta information attached to the request. When forwarded between peers,                                                                                                                   
// request meta headers are folded in matryoshka style.                                                                                                                                      
```
We add signatures to signatures while not providing any real benefit. Container nodes can do almost anything to data, they're explicitly allowed to, so an additional signature added to request doesn't change anything. Non-container nodes can only forward requests in which case an additional signature is useless as well, the original request needs to be checked.

The most critical thing is to do drop SN signatures. Then we can rely on TLS-only properties of the network to even allow 1:1 _unsigned_ requests.

roman-khimov:
Recent changes in NeoFS node make them irrelevant, we're always dealing with original requests and forwarding is done differently. Fixes #387.

### DIFF
--- a/session/types.proto
+++ b/session/types.proto
@@ -211,8 +211,7 @@ message XHeader {
   string value = 2 [json_name = "value"];
 }
 
-// Meta information attached to the request. When forwarded between peers,
-// request meta headers are folded in matryoshka style.
+// Meta information attached to the request (common things for all requests).
 message RequestMetaHeader {
   // Peer's API version used
   neo.fs.v2.refs.Version version = 1 [json_name = "version"];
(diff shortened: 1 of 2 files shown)


session/types.proto
cthulhu-rider, line 286: and this one?

roman-khimov: Fixed.

cthulhu-rider (approved):

carpawell (approved):

PR #404, opened 2026-07-27
session: clarify ResponseMetaHeader's fields presence

### LINKED ISSUE #397: Describe response meta header's goal and responsibility
It has true optional things (like X-Headers, as defined in their description), and at the same time it has things that do not look like optional helpers: epoch, and API version, but in practice they may not be set, and everything works.

I'm always frustrated when we have undefined behaviour.

One of:
1. make every field required and make it clear in the spec
2. make them fully optional, and make it clear in the spec that every request may define its meaning depending on the requirements this request has (e.g. https://github.com/nspcc-dev/neofs-api/pull/393)

\-

## Additional context
https://github.com/nspcc-dev/neofs-sdk-go/pull/822

carpawell:
Closes #397.

### DIFF
--- a/object/service.proto
+++ b/object/service.proto
@@ -354,7 +354,10 @@ message GetResponse {
   Body body = 1;
 
   // Carries response meta information. Header data is used only to regulate
-  // message transport and does not affect request execution.
+  // message transport and does not affect request execution. Non-status
+  // information can be attached only to stream's first response message and
+  // must be omitted for the subsequent ones. Error statuses are allowed to be
+  // transmitted in any message to terminate the stream.
   neo.fs.v2.session.ResponseMetaHeader meta_header = 2;
 
   // Carries response verification information. This header is used to
--- a/session/types.proto
+++ b/session/types.proto
@@ -245,18 +245,26 @@ message RequestMetaHeader {
   uint64 magic_number = 8 [json_name = "magicNumber"];
 }
 
-// Information about the response
+// Information about the response. For stream RPCs, non-status information
+// can be attached only to stream's first response message and must be
+// omitted for the subsequent ones. Error statuses are allowed to be
+// transmitted in any message to terminate the stream.
 message ResponseMetaHeader {
-  // Peer's API version used
+  // Server's API version used. Must be attached only if server's version is
+  // lower than client's one, optional otherwise.
   neo.fs.v2.refs.Version version = 1 [json_name = "version"];
 
-  // Peer's local epoch number
+  // Peer's local epoch number. Optional.
   uint64 epoch = 2 [json_name = "epoch"];
 
-  // Maximum number of intermediate nodes in the request route
+  // Maximum number of intermediate nodes in the request route.
+  //
+  // DEPRECATED: ignored since API v2.25.
   uint32 ttl = 3 [json_name = "ttl"];
 
-  // Response X-Headers
+  // Response X-Headers. Optional in general, but for certain requests it can
+  // be required for correct result processing, read method's description for
+  // information.
   repeated XHeader x_headers = 4 [json_name = "xHeaders"];
 
   // `ResponseMetaHeader` of the origin request.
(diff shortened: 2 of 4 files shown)


session/types.proto
```diff
@@ -247,24 +247,28 @@ message RequestMetaHeader {
 
 // Information about the response
 message ResponseMetaHeader {
-  // Peer's API version used
+  // Server's API version used. Must be attached only if server's version is
```
carpawell, line 253: > Must be attached

or maybe not must

roman-khimov: I think it should be there in this case. Client can adjust behavior if needed.

roman-khimov: "optional otherwise" still makes sense to add.

roman-khimov (dismissed):

session/types.proto
```diff
@@ -247,24 +247,28 @@ message RequestMetaHeader {
...
+  // lower than client's one, optional otherwise.
   neo.fs.v2.refs.Version version = 1 [json_name = "version"];
 
-  // Peer's local epoch number
+  // Peer's local epoch number. Optional.
   uint64 epoch = 2 [json_name = "epoch"];
 
-  // Maximum number of intermediate nodes in the request route
+  // Maximum number of intermediate nodes in the request route. Optional.
```
cthulhu-rider: i'd just mark it as deprecated. `origin` is already marked, and field affects nothing right now and completely senseless to me

session/types.proto
cthulhu-rider, line 253: what meaning could the field have in this case (server < client)? 

this is how i see it. If server detects unknown field in the request - it should be either rejected or forwarded. If all fields are known, behavior should be fixed and (minor)version-independent, making fields excessive again. Overall, I mean i dont expect having any "old way" to process fixed request that client should think about. This only complicates and can destabilize the system

and API version is available through node info rpc

session/types.proto
```diff
@@ -247,24 +247,28 @@ message RequestMetaHeader {
 
 // Information about the response
 message ResponseMetaHeader {
-  // Peer's API version used
+  // Server's API version used. Must be attached only if server's version is
+  // lower than client's one, optional otherwise.
   neo.fs.v2.refs.Version version = 1 [json_name = "version"];
 
-  // Peer's local epoch number
+  // Peer's local epoch number. Optional.
```
cthulhu-rider, line 257: - net info rpc returns it
- there is more reliable source for it - FS chain
- requirement for the field does not depend on the specific request. Server is free to do as it pleases

so the field is redundant for me. I don't see any point in continuing to support it

i think the same about `Optional` part of `x_headers`

cthulhu-rider (commented):
i'd drop as much fields as possible right now and keep only **really** handled ones. And describe **all** cases when they are set: there are few of them actually

i think it's also worth noting the behavior for server-side streams w/ multiple response messages like object GET. To me, only 1st message can have meta header

carpawell:
Added stream adjustments.

session/types.proto
carpawell: i dont mind. @roman-khimov

session/types.proto
carpawell: client may change its requests to the server accordingly

session/types.proto
carpawell: i think in general it can be used somewhere. there were even "caches" (that did not work) in SDK. i think once we are not dependent that much on epochs, we may drop it here too

session/types.proto
roman-khimov: Needs to be checked wrt real use. Reputation service cares about path for example. But if unused can be deprecated.

session/types.proto
roman-khimov: Can be deprecated in future. OK as optional for now.

session/types.proto
carpawell: no, seems they have never been used. nodes do not restrict anything based on this value. it is also meaningless to me without `origin` chain. the current reputation service is based on p2p local reviews and public keys

carpawell: deprecated

roman-khimov (dismissed):

object/service.proto
```diff
@@ -354,7 +354,9 @@ message GetResponse {
   Body body = 1;
 
   // Carries response meta information. Header data is used only to regulate
-  // message transport and does not affect request execution.
+  // message transport and does not affect request execution. It can be
+  // attached only to stream's first response message and must be omitted for
+  // the subsequent ones.
```
roman-khimov: What if error happens during object transmission, btw? In this case there is going to be an error only sent in meta without any body.

carpawell: made an exception

roman-khimov (approved):

PR #408, opened 2026-08-09
object: Drop meta headers from PUT chunk requests

cthulhu-rider:


roman-khimov (dismissed):
We can't verify data otherwise. And it's not the biggest problem of PUTs.

cthulhu-rider:
> We can't verify data otherwise. And it's not the biggest problem of PUTs.

data is verified via verification header

roman-khimov:
OK, got your point. Repeating meta doesn't make sense, that's true (changing it is also nonsense). So we can do this safely.

roman-khimov:
Thinking about it a bit more I find meta useful for replay attack protection. We currently don't have it, but we should. And the way it should work for chunks is to bind them to a particular stream which can be done using `nonce` in the first meta message and repeating it in all subsequent ones. In which case version field should be present too. And that makes meta header work as usual for chunks.

PR #409, opened 2026-08-12
session: add the new signature per request field

### LINKED ISSUE #406: Replace two verification signatures with a single one
I'm always frustrated when we're signing more things than actually needed. While we no longer need origin signature since #402 we still have _two_ other left and there are zero reasons for that.

API 2.26 should bring a new signature field that would sign serialized representation of _both_ body and meta header (sequentially). It can be a replacement for `RequestVerificationHeader` (preferred) or a field inside of it. Of course we're keeping compatibility so old header is still accepted and checked for old requests.

I don't have any.

carpawell:
Deprecate the old multi-signature way of request verification. Closes #406.

### DIFF
--- a/proto-docs/session.md
+++ b/proto-docs/session.md
@@ -205,14 +205,19 @@ Verification info for the request signed by all intermediate nodes.
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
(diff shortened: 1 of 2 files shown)


carpawell:
I feel extending the old verification header is more natural. But i can add a new signature field to every request instead, as stated in the issue, as a more preferable way.

roman-khimov:
I see, extending every request is no fun. If verification header is what they all have in common then let's use it.

roman-khimov (approved):

Issue #165, opened 2021-08-06
Add node attribute escape symbol note


realloc:
In nspcc-dev/neofs-node#700 we introduced the support for escape character in node attributes strings. This has to be reflected in API spec as well.

PR #410, opened 2026-08-19
Revert "[#167] netmap: Add attribute escaping description", fix #223

### LINKED ISSUE #223: Doubtful documentation of `NodeInfo.Attribute`
From `netmap.NodeInfo.Attribute` docs:
```
  // A string representation of the chain of attributes in NeoFS Storage
  // Node configuration uses ":" and "/" symbols, e.g.:
  //
  //        `NEOFS_NODE_ATTRIBUTE_1=key1:val1/key2:val2`
  //
  // Therefore the string attribute representation in the Node configuration must
  // use "\:", "\/" and "\\" escaped symbols if any of them appears in an attribute's
  // key or value.
```

The structure of the node attribute message provides needed feature of attribute inheritance. Mentioned doc part describes some sort of string interpretation for the particular storage node configuration. Do we need it in the protocol description? This seems like an implementation-dependent statements.

roman-khimov:
This reverts commit 2536e5a97b55556fdd4f5db9253c2e14cc9fb75a as inappropriate for API definition. This was a fix for https://github.com/nspcc-dev/neofs-api/issues/165 which leaked some implementation details into API for no real reason, key and value constraints are already defined in the spec and escaping or multi-key handling is out of scope. Also, we don't really have attribute chains, it's all flat KV space, so it's inaccurate and misleading as well.

### DIFF
--- a/netmap/types.proto
+++ b/netmap/types.proto
@@ -228,17 +228,6 @@ message NodeInfo {
   // `Attribute` is a Key-Value metadata pair. Key name must be a valid UTF-8
   // string (without zero bytes that are forbidden). Value can't be empty.
   //
-  // Attributes can be constructed into a chain of attributes: any attribute can
-  // have a parent attribute and a child attribute (except the first and the last
-  // one). A string representation of the chain of attributes in NeoFS Storage
-  // Node configuration uses ":" and "/" symbols, e.g.:
-  //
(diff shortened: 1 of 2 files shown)


carpawell (approved):

carpawell:
But it is kinda a change IMO. But i do not remember what is "chained attributes".

PR #411, opened 2026-08-19
*: clarify public keys, fix #211

### LINKED ISSUE #211: Clafiry binary format of the public keys
There are several fields of "binary public key" type, for example:

- [session_key](https://github.com/nspcc-dev/neofs-api/blob/431335054c3bd2fcff52226e04175b443bcb8a1e/refs/types.proto#L105) field of the `session.SessionToken` message
- [key](https://github.com/nspcc-dev/neofs-api/blob/431335054c3bd2fcff52226e04175b443bcb8a1e/refs/types.proto#L105) field in `refs.Signature` message

In current implementation the format is not specified, which on the one hand gives some flexibility to client applications, on the other hand makes decoding undefined.

I propose to clarify binary format of these fields (in `Signature` message we can specify per-`scheme` format).

roman-khimov:


refs/types.proto
```diff
@@ -133,8 +134,7 @@ enum SignatureScheme {
 
 // RFC 6979 signature.
 message SignatureRFC6979 {
-  // Public key used for signing. For N3 auth scheme, the field represents a
-  // verification script. The maximum allowed length is 1024 bytes.
```
carpawell, line 137: why was 1024 bytes restriction lost?

roman-khimov: It's not, `Signature` still mentions it and `SignatureRFC6979` references `Signature`.

carpawell (approved):

PR #418, opened 2026-09-01
Support streaming replication

cthulhu-rider:


object/service.proto
```diff
@@ -264,6 +264,9 @@ service ObjectService {
   // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER): \
   //   the container to which the replicated object is associated was not found.
   rpc Replicate(ReplicateRequest) returns (ReplicateResponse);
+
+  // TODO: docs.
+  rpc ReplicateV2(stream ReplicateRequest) returns (ReplicateResponse);
```
cthulhu-rider: my initial thought how to make new op as close to `Replicate` as possible is to stream its requests. Require additional messages to have `object.payload` (i.e. chunk) field only

having this, for new client to go back to old API, the only things to be done are
1. change method name to `Replicate`
2. place full object into the single message

2 will be needed anyway to implement a size-dependent logic transmitting small objects in one message

roman-khimov: Object size is known before the request is made, so I'd opt for simpler v2 protocol (two specific request-level fields for header and payload). Magic payload/non-payload handling is harder to implement to me. `ReplicateRequest` can be extended with payload field as well leaving object for complete object or header only depending on version. Needs to be checked against real code anyway.

cthulhu-rider: same as in #333, the main thing to decide is whether 1st message carries payload prefix (full payload in particular) or not

if no, we'll have `PutRequest`-like stream w/o meta/verification headers. Otherwise, prefix and extra chunk fields can be transmitted in the same field (currently proposed) or in the diff ones (PUT)

i'll make diff commits for all opts to make it clear

object/service.proto
```diff
@@ -264,6 +264,30 @@ service ObjectService {
   // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER): \
   //   the container to which the replicated object is associated was not found.
   rpc Replicate(ReplicateRequest) returns (ReplicateResponse);
+
+  // Save replica of the object on the NeoFS storage node. Both client and
+  // server must authenticate NeoFS storage nodes matching storage policy of
```
roman-khimov: ...be authenticated...?

object/service.proto
```diff
@@ -872,3 +896,47 @@ message ReplicateResponse {
   // `sign_object` flag set.
   bytes object_signature = 2;
 }
+
+// ReplicateV2 RPC request
+message ReplicateV2Request {
+  // Stream initialization data.
+  message Init {
+    // `object_id` field of the replicated object.
```
roman-khimov: I think we should mention that these fields specifically follow `object.Object` structure.

object/service.proto
```diff
@@ -872,3 +896,47 @@ message ReplicateResponse {
...
+    neo.fs.v2.refs.ObjectID object_id = 1;
+
+    // `signature` field of the replicated object.
+    neo.fs.v2.refs.Signature signature = 2;
+
+    // `header` field of the replicated object.
+    Header header = 3;
+
+    // Sender's signature of `object_id` field.
+    neo.fs.v2.refs.Signature node_signature = 4;
```
roman-khimov: So structurally I'd prefer to have these aux things separated from object. But I'm not sure if it's convenient to implement this way (more protobuf to handle).

roman-khimov (commented):
Can be merged as is and tuned before the release if needed. We want to minimize protobuf dances, ideally we should be able to copy header and payload into storage directly from messages, without any additional encoding.

object/service.proto
cthulhu-rider: spec says that this field is from the object

roman-khimov: I mean treating `Init` as `object.Object`, you can technically shuffle these fields somehow or omit some entirely (OID is known from header anyway). But the way they're specified now follow `object.Object` structure exactly (except for sig/sign).

object/service.proto
cthulhu-rider: had the same thought. Since structured field will require only a couple of extra bytes, i guess it'd be slightly easier to hande this way. So i'm for starting the same way as `ReplicateRequest` w/o payload

> I'm not sure if it's convenient to implement this way (more protobuf to handle)

from my exp handling single field is usually less tricky than grouping ones

object/service.proto
cthulhu-rider: ok

taking into account https://github.com/nspcc-dev/neofs-api/pull/418#discussion_r3966816527, this will be auto-resolved

object/service.proto
cthulhu-rider: fixed

roman-khimov (approved):

object/service.proto
```diff
@@ -872,3 +897,41 @@ message ReplicateResponse {
   // `sign_object` flag set.
   bytes object_signature = 2;
 }
+
+// ReplicateV2 RPC request
+message ReplicateV2Request {
+  // Stream initialization data.
+  message Init {
+    // Object to be replicated without payload.
+    neo.fs.v2.object.Object object = 1;
```
carpawell, line 906: will `header` be more understandable?

object/service.proto
```diff
@@ -264,6 +264,31 @@ service ObjectService {
...
+  // - **OK** (0, SECTION_SUCCESS): \
+  //   the object has been successfully replicated;
+  // - **INTERNAL_SERVER_ERROR** (1024, SECTION_FAILURE_COMMON): \
+  //   internal server error described in the text message;
+  // - **ACCESS_DENIED** (2048, SECTION_OBJECT): \
+  //   the client does not authenticate any NeoFS storage node matching storage
+  //   policy of the container referenced by the replicated object
+  // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER): \
+  //   the container to which the replicated object is associated was not found.
+  rpc ReplicateV2(stream ReplicateV2Request) returns (ReplicateV2Response);
```
carpawell, line 291: why "V2"?  i thought of it more as an alternative to the non-stream RPC

carpawell (approved):

object/service.proto
roman-khimov: Can be confused with `neo.fs.v2.object.Header`.

cthulhu-rider: considering that there is `header` field inside, unlikely

object/service.proto
cthulhu-rider: after #333 trial i think of it as a replacement more than an alternative. Although yeah, old API is not deprecated

cthulhu-rider: this can be discussed and changed if needed before release

roman-khimov: We can check it. 4K tests should be sensitive to these changes if they have any effect, so just use v2 for all requests and check the result, if it's the same then yes, v1 can be deprecated. If not, then v1 is here to stay for small objects.

PR #421, opened 2026-09-10
container: support container versioning

### LINKED ISSUE #380: Add container version field into object operation requests
See https://github.com/nspcc-dev/neofs-contract/issues/581 and https://github.com/nspcc-dev/neofs-s3-gw/issues/1244. With versions in-place we can have them passed in requests (optionally), then:
 * if there is no version, SN behaves in the same way as now
 * if there is one and SN has a different locally some (new type of) error is returned to caller

Then caller can detect it's out of date, re-request container and retry the operation using new data.

Obviously this implies versions don't change _often_ (theoretically retried request can get out of sync as well), but for things stored in container currently that's true.

carpawell:
Closes #380.


container/service.proto
```diff
@@ -277,6 +289,10 @@ message GetResponse {
 
     // Session token if the container has been created within the session
     neo.fs.v2.session.SessionToken session_token = 3;
+
+    // Version of the container above. Starting from API v2.27.0, containers
+    // are versioned and changing container increments its version.
+    uint64 container_version = 4;
```
carpawell, line 295: it is the only way to get container version. mb it can be part of the Container message

container/service.proto
```diff
@@ -91,6 +94,9 @@ service ContainerService {
   //   container not found;
   // - **EACL_NOT_FOUND** (3073, SECTION_CONTAINER): \
   //   eACL table not found.
+  // - **CONTAINER_CHANGED** (3076, SECTION_CONTAINER): \
```
roman-khimov, line 97: Doubt it makes any difference for getters. They just return the latest known data.

container/service.proto
roman-khimov: To me it's a part of `Container`.

container/service.proto
```diff
@@ -336,7 +352,14 @@ message ListResponse {
...
 
-// Set Extended ACL
+// Set Extended ACL.
+//
+// Starting from API v2.27.0, requester can attach container version to
+// `meta_header.x_headers` to ensure container state is up to date. If server's
+// known version for container is higher, it must return *CONTAINER_CHANGED**
+// (3076) response status with no payload. Header format:
+//   Key: __NEOFS__CONTAINER_VERSION
+//   Value: <base-10 positive integer version counter with no leading zeros>.
```
roman-khimov, line 362: We need a single description of this X-Header somehow and then just a reference here like "behaviour can be augmented with __NEOFS__CONTAINER_VERSION x-header"

object/service.proto
```diff
@@ -45,6 +45,9 @@ service ObjectService {
   //   object container not found;
   // - **TOKEN_EXPIRED** (4097, SECTION_SESSION): \
   //   provided session token has expired.
+  // - **CONTAINER_CHANGED** (3076, SECTION_CONTAINER): \
```
roman-khimov, line 48: Also makes no sense for getter, there is an OID, there is its data, it's immutable.

object/service.proto
```diff
@@ -139,6 +148,9 @@ service ObjectService {
   //   the requested object has been marked as deleted;
   // - **CONTAINER_NOT_FOUND** (3072, SECTION_CONTAINER): \
   //   object container not found;
+  // - **CONTAINER_CHANGED** (3076, SECTION_CONTAINER): \
+  //   if requester attached container version he knows and it is lower than
+  //   the server's one.
```
roman-khimov, line 153: Same here

status/types.proto
```diff
@@ -158,6 +158,11 @@ enum Container {
 
   // [**3075**] Async container operation timed out.
   CONTAINER_AWAIT_TIMEOUT = 3;
+
+  // [**3076**] Requested operation with container is outdated: container
+  // has been changed. Can be returned only if requester provides the latest
+  // container version, see requests documentation.
+  CONTAINER_CHANGED = 4;
```
roman-khimov, line 165: OUTDATED_CONTAINER?

container/service.proto
carpawell: why? user calling for access info, cant it be useful to know container has changed?

status/types.proto
carpawell: dont mind

object/service.proto
carpawell: i dont know why it is needed at all then. what is the difference b/w PUT and GET? i thought access is the key

container/service.proto
carpawell: done, but no difference to me: still need many places to add info about this header

container/service.proto
carpawell: moved to Container, but I am not confident with this field. to me, it should be a read-only property from Get requests

proto-docs/session.md
```diff
@@ -396,6 +396,13 @@ affect system behaviour:
   how many past epochs the node can look up through. The `value` is string
   encoded `uint64` in decimal presentation. If set to '0' or not set, only the
   current epoch will be used. DEPRECATED: header ignored by servers.
+* __NEOFS__CONTAINER_VERSION \
+  Starting from API v2.27.0 and for a limited set of requests only, requester
+  can attach container version to extended headers to ensure container state
+  is up to date. If server's known version for container is higher, it must
```
cthulhu-rider: OK if `lower`? Thinking of client specifying 42 while current is 1

carpawell: hm, it is interesting. if client knows more about the world, should we return `OUTDATED_CONTAINER` error?
@roman-khimov, @smallhive

carpawell:
Dropped from `GetExtendedACL`, `Get`, `Head`.

proto-docs/session.md
roman-khimov: I'd expect an exact match, so yeah, it can be like `CONTAINER_VERSION_MISMATCH`.

container/service.proto
```diff
@@ -382,7 +392,8 @@ message SetExtendedACLResponse {
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
 }
 
-// Get Extended ACL
+// Get Extended ACL.
+// Behaviour can be augmented with __NEOFS__CONTAINER_REVISION x-header.
```
cthulhu-rider: seems to be a leftover https://github.com/nspcc-dev/neofs-api/pull/421#issuecomment-5663163877

container/service.proto
```diff
@@ -382,7 +392,7 @@ message SetExtendedACLResponse {
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
 }
 
-// Get Extended ACL
+// Get Extended ACL.
```
cthulhu-rider, line 395: unrelated

cthulhu-rider (dismissed):

container/types.proto
```diff
@@ -116,4 +116,17 @@ message Container {
...
+
+  // Container revision. It increments every time container's properties are
+  // changed by the owner (or the owner itself is changed).
+  //
+  // Do not confuse it with API version: this field describes how many times
+  // container has been changed since its creation, while API version describes
+  // proto message format.
+  //
+  // It must only be set by storage nodes and must not be filled on the client
+  // side. The initial revision after a successful container creation call is 1.
```
roman-khimov, line 128: Why not zero?

object/service.proto
```diff
@@ -503,7 +514,7 @@ message DeleteResponse {
   neo.fs.v2.session.ResponseVerificationHeader verify_header = 3;
 }
 
-// Object HEAD request
+// Object HEAD request.
```
roman-khimov, line 517: :eyes:

container/types.proto
carpawell: what does "zero revision" term mean? also, i would be easier to treat mistakes as an error, missing field will be found immediately. fixed to merge PR faster

cthulhu-rider (approved):

roman-khimov (approved):

PR #424, opened 2026-09-17
*: drop assertion regarding the immutability of request execution

carpawell:
Meta headers do change request results based on X-Headers at least: #421, #345.


carpawell:
We also have the same statement in `*Response` messages. Can be fixed also, but it is not currently violated.

roman-khimov (approved):

PR #425, opened 2026-09-17
container: make [Set/Remove]Attribute request follow common pattern

carpawell:
They do not have meta header, so they are not as extendable as other requests. In particular, it does not allow implement #421 without stumbling.


container/service.proto
```diff
@@ -563,13 +570,36 @@ message SetAttributeRequest {
...
+  // authenticate nodes and check the correctness of transmission.
+  neo.fs.v2.session.RequestVerificationHeader verify_header = 4;
 }
 
 // Attribute setting response
 message SetAttributeResponse {
   // Operation execution status.
+  // DEPRECATED: use more common response meta information approach:
+  // metadata header below. Servers must attach status to this field
+  // for clients with API version <2.27.0.
```
carpawell, line 590: hm, but basically it means for any request with no API version, since there was no meta header and API version before this PR

container/service.proto
```diff
@@ -563,13 +570,36 @@ message SetAttributeRequest {
   Body body = 1;
 
   // Signature of stable-marshalled `body` field.
+  // DEPRECATED: use more common request meta information approach:
+  // metadata and verification headers below.
   neo.fs.v2.refs.Signature body_signature = 2;
+
+  // Carries request meta information.
+  neo.fs.v2.session.RequestMetaHeader meta_header = 3;
```
cthulhu-rider, line 578: i'm not for this field at all

but having it, each field inside must be documented. For example, TTL must not be set. Bearer token too. All deprecated fields because this field appears after existing deprecations. `body.session_token(_v1)` already exists, so we must doc the combination

same for `verify_header`

container/service.proto
```diff
@@ -563,13 +570,36 @@ message SetAttributeRequest {
...
+
+  // Carries request verification information. This header is used to
+  // authenticate nodes and check the correctness of transmission.
+  neo.fs.v2.session.RequestVerificationHeader verify_header = 4;
```
cthulhu-rider, line 582: it will contain single signature. There is `body_signature` field already but we are deprecating it. Just weird to me

container/service.proto
```diff
@@ -563,13 +570,36 @@ message SetAttributeRequest {
...
+  // `SetAttributeResponse` has an empty body. Operation status is attached to
+  // the meta header.
+  message Body {}
+
+  // Body of set attribute response message.
+  Body body = 2;
+
+  // Carries response meta information. Header data is used only to regulate
+  // message transport and does not affect request execution.
+  neo.fs.v2.session.ResponseMetaHeader meta_header = 3;
```
cthulhu-rider, line 602: it just deepens the status, nothing more. And current behavior of other fields must be specified (all prohibited except status)

cthulhu-rider (changes requested):
> they are not as extendable as other requests

this is incorrect. Fields can be added to `body`

> In particular, it does not allow implement #421 without stumbling

extend `body` with `refs.Version api_version` + `uint64 revision` (ok, it could be `repeated session.XHeader x_headers` instead, but it's too much for me, protobuf is a typed format) fields and that's it

in total, i think we only add deprecations and complicate the structure instead of focusing on the real need: add 1-2 new fields

carpawell:
> this is incorrect. Fields can be added to body

I meant we need a PR to solve my problem. A simple [doc line](https://github.com/nspcc-dev/neofs-api/blob/1fe755e0848eb32ee22dde5d67bf954db32c656d/container/service.proto#L506) did not work. Sooner or later, smth will be done with the meta header, and these requests will be forgotten again.

If we do not want to attach excessive fields, i can suggest MetaHeaderV2/VerifyHeaderV2 and add it only to the `*Attribute` calls. Later, it can also be the only MetaHeader in other requests, with deprecation of the old ones.

cthulhu-rider:
> smth will be done with the meta header, and these requests will be forgotten again

no matter whether new field is added to cross-request struct or not, behavior (doc) of each op affected within current task should be updated separately. Like for container revision: it was necessary to write on each request whether the header is applicable to it or not, while structurally it was added to all requests. Same thing w/ TTL, etc.

and in general, adding a requests' payload to shared header (especially as untyped KV one) always seemed like a bad approach to me and continues to be. Protobuf is not about this. This is why #424 was needed: in fact, the request payload is now passed in the meta header instead of body. This is illogical to me. In Set/RemoveAttribute, it was done intentionally differently

roman-khimov:
While originally `RequestMetaHeader` was intended to be used for transport regulation it has evolved since to provide more functionality. Most of this functionality is not applicable to `[Set/Remove]Attribute` and they were created using a different approach. Now this duality is a problem for sure, we need some common principles our API design follows. To get it right let's answer a set of questions here.

1. Is there any data that is common, applicable to all requests?
I think so. API version, session tokens, magic number and (potentially) nonce are useful for all requests.
2. Is there any data in `RequestMetaHeader` that is only relevant for some requests and not others?
Yes, TTL is only applicable to object service, as well as bearer token.
3. What's the meaning of X-Headers?
There are not a lot of traces of it other than f059c49dec5561166abe14a87c9ee469eade60c1. But there are also `__NEOFS__NETMAP_EPOCH`, `__NEOFS__NETMAP_LOOKUP_DEPTH`, so it looks like it was designed to augment requests in a generic way (defeating the transport-only meta header purpose, btw). This makes some sense when we have a set of requests that can be processed slightly differently depending on something. In that sense it's somewhat similar to TTL or bearer token. Now it has grown to include `__NEOFS__EC_RULE_IDX`, `__NEOFS__EC_PART_IDX` and `__NEOFS__CONTAINER_REVISION`. It's also important to note that these headers can be used in EACL filters.
4. Should we use common messages for common cross-request things?
Yeah, why not? Duplicating the same sets of fields over and over again doesn't seem productive and makes maintaining them harder.
5. Can we have some fields in common structures that are applicable to a number of requests, but not all of them?
This better be avoided.
6. What's better, body+sig or body+meta+sig?
The original design with meta _mostly_ handling transport made and matryoshkas running around the question rather obvious (refs. 4b1eb63b59b9ca92e9d90bdcf7c0d3692c1a0082), meta had to be managed/signed separately. Now with #409 implemented it leads to some questions that take time to explain (like the exact way of body+meta concatenation to derive a single signature). At the same time body+meta+sig has some cons too, this way cross-request meta is ensured on a different level (like _any_ request has an API version, without even looking into its body) and common code can rely on this structure.
7. Can we have body-only requests?
Doesn't look like, signatures need to be calculated over some data that can be easily identified (not "a subset of fields not including...").
8. Should we change all of our requests to some new scheme, like body+sig?
Technically possible, but I doubt it improves anything for the client.

This PR makes `[Set/Remove]Attribute` follow the current common request scheme. It at the same time breaks internal consistency of these requests (duplicated session tokens at least).

Extending their bodies with API/x-headers deepens the difference between these requests and others, meaning an additional code would be required to handle them and more effort would be required to maintain any common behavior (extend/drop something for all requests) if needed.

Overall, _any_ common structure is stronger to me. We can think of improving it, we can think of easing the pain of internal request inconsistency (up to making v2 requests), but otherwise we better follow some common scheme. We do have an exception in `Replicate`, but that one is internal at least.

cthulhu-rider:
> meaning an additional code would be required to handle them and more effort would be required to maintain any common behavior (extend/drop something for all requests) if needed

that's it. Ops' behavior is rarely common and never exactly the same. Therefore, "just add/change field in shared message and that's all" is an illusion, because of which the lack of strictness and abstractions in ops only increases. For example:
1. session tokens may not be applicable to all requests
2. those that are applicable have different requirements (like verbs)

while 2 can be explained in docs, 1 is just "this field exists, but it cannot be set"

i also see adding opts to raw KV headers as a growing pit, further reducing the point of using protobuf in general

yes, there would be a need to press Ctrl-C Ctrl-V up to ~5 times sometimes. But being backed this up with per-case detailed docs, in my opinion, everything would be done much more clearly and controllably, and there would be fewer cross-connections and implicitities. Current practice shows that when changing 1-5 queries, no one thinks about the other 10 ones

> What's better, body+sig or body+meta+sig?

to me it's body with API version inside + sig (or version+body+sig)

> we can think of easing the pain of internal request inconsistency (up to making v2 requests)

this is the only compromise I can see so far. Otherwise, I simply don't see the point in perfoming surgery of two ops that worked perfectly fine before. And when two fields are needed for a specific task: what worked perfectly is deprecated, unusable fields are added (incl. already deprecated ones)

to me only some personal perceptions are addressed. It's also worth considering the clients who will ultimately be forced to follow the new, more bloated format. At least docs for all new fields (incl. each nested one) should be as complete as possible. I'm not even talking about server implementation with backwards compatibility: it will be much more complex than 2 new fields

to sum up, i still don't see the point in proposed approach. Imo the efforts being made are disconnected from particular need. I suggest to do the bare minimum to get revisioned container attributes and move standardization to a separate issue

roman-khimov:
These two calls better be resolved with V2, both request and reply structures can be made stronger reusing existing meta headers. Then we should prepare for some hardening of our requests, so let's start it with these ones (it will affect contracts):
1. Body/meta/sig common structure, tokens retrieved from meta, body only containing request-specific things.
2. No separate signatures for parameters/body.
3. Move valid_until right into request meta header first-class, it's applicable to any request.
4. Adjust contracts to accept `Call(data1, data2, originalRequest struct{body, meta, sig []byte})`, verify FS requests (and data accuracy) in IR.

Then a similar structure will be applied to all other requests.

PR #426, opened 2026-09-17
object: document __NEOFS__NONCE, make it a bit special

roman-khimov:
This attribute is used by the S3 gateway now and it contains some bits of randomness. It's absolutely useless for SEARCH requests in this way, so nodes can skip indexing it.


object/types.proto
```diff
@@ -219,6 +219,11 @@ message Header {
   //   is the only way to delete/lock objects. It MUST be a single stringified
   //   (according to [refs.ObjectID] message) object ID with no leading or
   //   trailing spaces.
+  // * __NEOFS__NONCE \
+  //   Can contain any data and is used to prevent OID collisions for otherwise
+  //   identical objects (same headers, same payload). Usually it contains 8
+  //   base64-encoded bytes of randomness. This attribute can not be used in
+  //   SEARCH requests since it's not indexed by nodes.
```
cthulhu-rider, line 230: also worth writing about this in search docs

roman-khimov: Had this thought, but it's a bit special since it's neither deprecated, nor special (`$Object:`). Added anyway.

cthulhu-rider (approved):

